PR #1, opened 2020-02-27
Setup frontend app

MartinForre:


eskil0312:
Why is there one ./frontend and on ./src/frontend ?

PR #6, opened 2020-02-28
Actual mpp structure and PersonnelReques models

eskil0312:
Added PersonnelRequest and related models


eskil0312:
@MartinForre  @olerichard What do you think of the model names? Are they a bit general? Should we prefix with f.eks "Resource"?

PR #7, opened 2020-02-28
Basic allocate contract form

MartinForre:


src/frontend/src/index.tsx
```diff
@@ -1,18 +1,31 @@
...
 import LandingPage from './pages/LandingPage';
-// import ContractsOverviewPage from './pages/ContractsOverviewPage';
-// import ContractDetailsPage from './pages/ContractDetailsPage';
 import ProjectPage from './pages/ProjectPage';
+import ApiClient from './api/ApiClient';
+import AppContext from './appContext';
 
 const App: React.FC = () => {
+    const fusionContext = useFusionContext();
+    const apiClient = new ApiClient(
```
eskil0312: Memorize client?

MartinForre: Shouldn't be that many rerenders of the app component, but sure we can memoize the client

eskil0312: Don't bother

PR #9, opened 2020-02-28
Active request page

eskil0312:


src/frontend/src/models/PersonnelRequest.ts
```diff
@@ -5,7 +5,13 @@ import Position from './Position';
 import Project from './Project';
 import Comment from './Comment';
 
-export type RequestState = 'Created' | 'Submitted' | 'Approved' | 'Rejected' | 'Provisioned';
+export enum RequestState {
```
MartinForre: Why was this changed to an enum?

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ActiveRequestsPage/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const [isFetching, setIsFetching] = React.useState<boolean>(false);
+    const [error, setError] = React.useState(null);
+    const [selectedRequests, setSelectedRequests] = React.useState<PersonnelRequest[]>([]);
+    const { apiClient } = useAppContext();
+    const getRequestsAsync = async () => {
+        setIsFetching(true);
+        setError(null);
+        try {
+            const response = await apiClient.getPersonnelRequestsAsync('123', '123'); //TESTING VALUES
+            const activeRequests = response.data.value.filter(
```
MartinForre: This should probably be handled by the backend with an OData filter

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ActiveRequestsPage/columns.ts
```diff
@@ -0,0 +1,40 @@
...
+        sortable: true,
+    },
+    {
+        accessor: request => request.position?.name || '',
+        key: 'position',
+        label: 'Position',
+        sortable: true,
+    },
+    {
+        accessor: request => request.position?.taskOwner?.positionId || '',
```
MartinForre: Should this use a position card/column?

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/components/RequestStateFlow.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+const states = ["Created state", "Sumbitted state", "Approved state"]
+
+const getRequestStates = (request: PersonnelRequest): RequestItemState[] => {
+    const state = request.state;
+
+    switch (+state) {
+        case 0:
+            return ["approved", "pending"];
```
MartinForre: I think we should return the same number of steps no matter what the state is. Otherwise it might look like different workflows for the request with different amount of steps

src/frontend/src/models/PersonnelRequest.ts
eskil0312: As of right now the backend returns a number between 0 and 3 for status, but this is maybe a bug?

MartinForre: @HansDahle Could the request state be the string value of the enum, instead of int? Easier for both the resource app and other consumers to reason about a human readable string value vs just a random number

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ActiveRequestsPage/index.tsx
eskil0312: It should, but as of right now it's not implemented. Should I just remove the frontend filter?

MartinForre: The backend just returns bogus data. We should set up the frondend as if the backend was implemented correctly

eskil0312: Changed to OData filter in url

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ActiveRequestsPage/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const [activeRequests, setActiveRequests] = React.useState<PersonnelRequest[]>([]);
+    const [isFetching, setIsFetching] = React.useState<boolean>(false);
+    const [error, setError] = React.useState(null);
+    const [selectedRequests, setSelectedRequests] = React.useState<PersonnelRequest[]>([]);
+    const { apiClient } = useAppContext();
+    const getRequestsAsync = async () => {
+        setIsFetching(true);
+        setError(null);
+        try {
+            const response = await apiClient.getPersonnelRequestsAsync('123', '123', "RequestState eq 0 or RequestState eq 1"); //TESTING VALUES
```
MartinForre: Would like the OData part of the query to be handled by the client. The active requests page shouldn't control the query string parameters in requests made by the api client. And also the $filter should be simpler to read. E.g. 'isActive eq true'. 'RequestState eq 0 or RequestState eq 0' will mean nothing for the next developer that comes across this piece of code.

Another thing, in general about the api client: we should probably unwrap the responses within the api client instead of returning the HttpResponse<T> type 🤔 If you change the methods you've introduced in this PR, me and @olerichard will do the same

src/frontend/src/models/PersonnelRequest.ts
HansDahle: yes, guess a converter is forgotten

src/frontend/src/api/ApiClient.ts
```diff
@@ -33,11 +33,17 @@ export default class ApiClient {
         return this.httpClient.getAsync<ApiCollection<Personnel>, FusionApiHttpErrorResponse>(url);
     }
 
-    async getPersonnelRequestsAsync(projectId: string, contractId: string, filter?: string) {
+    async getPersonnelRequestsAsync(
```
eskil0312: @MartinForre  Something like this?

MartinForre: Yes :+1:

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ActiveRequestsPage/columns.ts
eskil0312: @MartinForre Moved` PositionColumn` component from `./src/pages/ProjectPage/pages/ContractsOveriewPage/`components to `./src/pages/ProjectPage/components` and used that

src/frontend/src/models/PersonnelRequest.ts
eskil0312: Changed back to type

PR #14, opened 2020-03-04
Manage Personnel Page

olerichard:


src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/PersonnelColumns.tsx
```diff
@@ -0,0 +1,67 @@
...
+    sortable: true,
+  },
+  {
+    key: 'Phone',
+    accessor: 'phoneNumber',
+    label: 'Phone Number',
+    priority: 15,
+    sortable: true,
+  },
+  {
```
olerichard: These two are not available in the api when getting personnel. 
placeholder value for now

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/PersonnelColumns.tsx
```diff
@@ -0,0 +1,67 @@
...
+    priority: 20,
+    sortable: true,
+  },
+  {
+    key: 'positions',
+    accessor: r => "???",
+    label: 'Positions',
+    priority: 25,
+    sortable: true,
+  }
```
olerichard: These two are not in the api when fetching personnel for now. 
Placeholder values

src/frontend/src/pages/ProjectPage/index.tsx
```diff
@@ -4,13 +4,15 @@ import ContractsOverviewPage from './pages/ContractsOverviewPage';
...
 import ScopedSwitch from '../../components/ScopedSwitch';
+import ManagePersonellPage from './pages/ContractPage/pages/ManagePersonnelPage';
 
 const ProjectPage = () => {
     return (
         <ScopedSwitch>
             <Route path="/" exact component={ContractsOverviewPage} />
             <Route path="/allocate" exact component={AllocateContractPage} />
             <Route path="/:contractId" component={ContractPage} />
+            <Route path="/:contractId/managedpersonnel" component={ManagePersonellPage} />
```
MartinForre: This should probably be in the ContractPage, not project page to get the navigation? Also: /managedpersonnel or /managepersonnel?

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/AddPersonnelSideSheet/AddPersonnelFormHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+type HeaderProps = {
+  headers: string[];
+}
+
+const AddPersonnelFormHeader : React.FC<HeaderProps> = ({headers}) => {
+
+  const cellClassName = classNames(styles.cell, styles.header);
+  return (
+    <>
+      <div className={classNames(cellClassName, styles.expand)} >
```
MartinForre: This could be combined with the classNames on line 11

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/AddPersonnelSideSheet/AddPersonnelFormHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+        {headers.map((headerlabel) => (
+          <div style={{flexGrow:1}}
```
MartinForre: flexGrow could be moved to a css class

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/AddPersonnelSideSheet/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    <ModalSideSheet
+      header="Add Person"
+      show={isOpen}
+      onClose={() => {
+        setIsOpen(false);
+      }}
+      safeClose
+      safeCloseTitle={`Close Add Person? Unsaved changes will be lost.`}
+      headerIcons={[
+        <Button key={"AddPerson"} outlined onClick = {()=> {setFormState([...formState,{personnelId:uuid(), name:"",phoneNumber:"",mail:"",jobTitle:""}])}} > + Add Person </Button>,
```
MartinForre: the on click handler should be moved to a const onAddPerson = React.useCallback(...)

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/AddPersonnelSideSheet/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <Button disabled={!(isFormDirty && isFormValid)} key={"save"} outlined onClick = {()=> {savePersonnelChangesAsync()} }> Create </Button>
```
MartinForre: No need to wrap savePersonnelChangesAsync in a function. Could be passed directly to the button onClick={savePersonnelChangesAsync}

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/AddPersonnelSideSheet/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+      ]}
+      isResizable
+      minWidth={640}
+    >
+      {saveInProgress && <Spinner centered  size={100} title={"Saving Personnel"} />  }
+      {!saveInProgress && isOpen &&  <div className={containerClassNames}>
```
MartinForre: Do we need to hide the entire table when saving? Wouldn't it be better for the user to just show a spinner in the save button or whatever while saving instead of switching the entire UI?

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/AddPersonnelSideSheet/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <div className={styles.table}
+            style={{ gridTemplateColumns: columnTemplate, gridTemplateRows: rowTemplate}}
+        >
+          <Header headers={headers} />
+          <table>
+            <tbody>
+              {
+                formState.map((person) => (
+                  <tr key={`person${person.personnelId}`} style={{marginBottom:"16px", display:"flex"}}>
```
MartinForre: All of the styling here gould be moved to css classes

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/hooks/usePersonnel.ts
```diff
@@ -0,0 +1,42 @@
...
+import { useAppContext } from '../../../../../../../appContext';
+
+const usePersonnel = (contractId?: string,projectId?:string) => {
+    const [personnel, setPersonnel] = React.useState<Personnel[]>([]);
+    const [isFetchingPersonnel, setIsFetchingPersonnel] = React.useState(false);
+    const [personnelError, setPersonnelError] = React.useState<Error | null>(null);
+    const { apiClient } = useAppContext();
+       
+    const fetchPersonnel = async (contract: string,project:string) => {
+        setIsFetchingPersonnel(true);
```
MartinForre: Should reset the personnelError to null here

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/AddPersonnelSideSheet/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+      return  
+    
+    setSaveInProgress(true);
+    await Promise.all(formState
+      .map(async person => await apiClient.updatePersonnelAsync(currentContext.id,contractId,person)))
+      .then(() => {
+        setSaveInProgress(false)
+        setIsOpen(false);
+      })
+      .catch(() => setSaveInProgress(false))
```
eskil0312: Should probably let the use know an error occurred and the personnel might not have been saved

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/AddPersonnelSideSheet/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  
+  return(
+    <ModalSideSheet
+      header="Add Person"
+      show={isOpen}
+      onClose={() => {
+        setIsOpen(false);
+      }}
+      safeClose
+      safeCloseTitle={`Close Add Person? Unsaved changes will be lost.`}
```
eskil0312: Should probably also add safe close cancel/confirm label to increase clarification. Something like this ` safeCloseCancelLabel ="Continue editing"` `safeCloseConfirmLabel = "Discard current changes"`

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/PersonnelColumns.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { DataTableColumn, useTooltipRef, DoneIcon, WarningIcon, CloseIcon } from '@equinor/fusion-components';
+import Personnel from '../../../../../../models/Personnel';
+
+export type DataItemProps = {
+  item: Personnel;
+};
+
+// TODO: Get proper icons
+const AdStatus = {
+  Available : {text: "Azure AD Approved",icon:<DoneIcon color={"green"}/>},
```
eskil0312: Use equinor colors?

olerichard: they are just placeholders till i get the real icons.

src/frontend/src/pages/ProjectPage/pages/ContractPage/index.tsx
```diff
@@ -86,7 +86,7 @@ const ContractPage: React.FC<ContractPageProps> = ({ match }) => {
                     <div className={styles.details}>
                         <ScopedSwitch>
                             <Route path="/" exact component={ContractDetailsPage} />
-                            <Route path="/personnel" component={ManagePersonellPage} />
+                            <Route path="/managepersonnel" component={ManagePersonellPage} />
```
MartinForre: Since the rest of the routes uses dashes to seperate words, we should probably keep doing that

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/AddPersonnelSideSheet/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+                    >
+                        <Header headers={headers} />
+                        <table>
+                            <tbody>
+                                {formState.map(person => (
+                                    <tr
+                                        key={`person${person.personnelId}`}
+                                        className={styles.tableRow}
+                                    >
+                                        <td style={{ flexGrow: 1 }}>
```
MartinForre: flexGrow should be in the less file

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/PersonnelColumns.tsx
MartinForre: import { styling } from "@equinor/fusion-components";
styling.colors.green, styling.colors.orange, styling.colors.red <- will give you the equinor/eds hue's

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/hooks/useAddPersonnelForm.ts
```diff
@@ -0,0 +1,29 @@
...
+    name:"",
+    phoneNumber:"",
+    mail:"",
+    jobTitle:""
+  }]);
+
+const useAddPersonnelForm = (defaultState?: Person[] | null) => {
+    const validateForm = useCallback((formState: Person[]) => {
+        return formState
+        .find(p =>
```
eskil0312: Isn't this the same as: `.find(p => p.name && p.phoneNumber && p.mail)`?  Would be easier to understand

MartinForre: `p.name && p.phoneNumber && p.mail` would return something like `boolean | undefined` or `boolean | null` while wrapping the expression in `Boolean(...)` would ensure the boolean type and make it easier to use (especially with TypeScript)

eskil0312: It would be easier to use, but we are not using it so it won't matter. The result from the `find()` would be the same either way. We could wrap the expression in a `Boolean(...)` or `!!(...)`, but that would only affect the readability.  But is the `Boolean(...) === false ? false : true` needed or could we just use `Boolean(...)?`

MartinForre: The ternary is not necessary

olerichard: Find does not return a boolean. it returns the object. 
Find will here iterate until it finds a invalid line and then return it. 
Hence the ternery at the end. it equates the result from find to a boolean. 

If something is found, the form is invalid, else its valid.  
guess I could change it to;
`return  !formState.find(p => Boolean(p.name && p.phoneNumber && p.mail) === false) `

eskil0312: If you want to return a boolean I would suggest using `some()` instead. `some() `returnes a boolean based on if a statement is true or false instead of a object like `find()` does. And if the intention is just want to check if a line is invalid, the ternary at the end is by my opinion not needed. Wouldn't something like this work? `formState.some(p => !(p.name && p.phoneNumber && p.mail))`

eskil0312: Note: I now see that the ` ? false : true` ternary is **outside** the `find()`, my bad 😅. But my suggestion above still stands

olerichard: wasnt aware some() returned a bool on first hit. So that makes sense. 
The false check is still needed though in the end is still needed though. 
Cause I need to return false if something is found.

PR #15, opened 2020-03-04
Filter for Actual MPP Page and Active Requests Page

eskil0312:


MartinForre:
@eskil0312 Rather that filtering on base position, should we filter on the discipline for that base position? I think that would make more sense to the users

PR #22, opened 2020-03-06
Manage personnel - First- and Lastname split

olerichard:
Split "Person" column into First and last name. 
Both in overview and on the edit sheet. 

Stole styling from the other pages and added to manage personnel. 
This to assure consistent padding etc between the pages. 

Add personnelSideSheet now uses fixed fullscreen size from the sidesheet componenet. 

Removed a bunch of unused Css that was copied inn when in from DataTable. 
Only the parts that are actually in use should be left.


src/frontend/src/models/Personnel.ts
```diff
@@ -6,8 +6,8 @@ type Personnel = {
     personnelId: string;
     azureUniquePersonId?: string;
     name: string;
-    firstName?: string;
-    lastName?: string;
+    firstName: string;
+    lastName: string;
```
MartinForre: Since `firstName `and `lastName `on `Person `is `string | null`, `firstName `and `lastName `on `Personnel `should also be `string | null`. Also. Strings in C# are nullable, så all strings we receive from the API might be null and we should handle those cases

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/AddPersonnelSideSheet/AddPersonnelFormHeader.tsx
```diff
@@ -6,10 +6,10 @@ type HeaderProps = {
   headers: string[];
 }
 
-const AddPersonnelFormHeader : React.FC<HeaderProps> = ({headers}) => {
+const AddPersonnelFormHeader: React.FC<HeaderProps> = ({ headers }) => {
   return (
     <>
-      <div className={classNames(styles.cell, styles.header, styles.expand)} >
+      <div className={classNames(styles.cell, styles.header)} >
```
MartinForre: I think the entire table would be simpler if we used `<thead><tr><th>{headerLabel}</th><th>{headerLabel}</th><th>{headerLabel}</th></tr></thead>` instead of combining divs and a table

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/AddPersonnelSideSheet/index.tsx
```diff
@@ -83,11 +83,19 @@ const AddPersonnelSideSheet: React.FC<AddPersonnelToSideSheetProps> = ({
...
+                lastName: "",
+                phoneNumber: "",
+                mail: "",
+                jobTitle: ""
+            },
         ]);
     }, [formState]);
 
-    const headers = ['Name', 'Mail', 'Phone'];
+    const headers = ['Firstname', 'Lastname', 'Mail', 'Phone'];
```
MartinForre: 'First name' and 'Last name'?

PR #24, opened 2020-03-06
Fixed saving and creating users. disabled email edit for exisiting users

olerichard:
In Api client:
updatePersonnelAsync is now put and takes in person paramter.
New createPersonnelAsync created to handle post request.
resurceCollections updated to reflect this.

Models:
added created to Person model. Need this to differentiate between a locally created user and one from the api.
created and updated fields also added to Personnel.

PersonnelSide sheet and saving routine: 
When saving personnel it now checks if user is created to select correct endpoint.
When editing users email field is disabled for already created users.
Added placeholder to input fields, just so it still show data even if its disabled.

### DIFF
--- a/src/frontend/src/api/ResourceCollection.ts
+++ b/src/frontend/src/api/ResourceCollection.ts
@@ -22,17 +22,14 @@ export default class ResourceCollection {
         return combineUrls(this.contracts(projectId), contractId);
     }
 
-    personnel(projectId: string, contractId: string): string {
-        return combineUrls(this.contract(projectId, contractId), 'resources', 'personnel');
+    personnel(projectId: string, contractId: string, personnelId?: string): string {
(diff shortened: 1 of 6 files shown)


src/frontend/src/api/ResourceCollection.ts
```diff
@@ -30,6 +30,10 @@ export default class ResourceCollection {
         return combineUrls(this.contract(projectId, contractId), 'resources', 'personnel-collection');
     }
 
+    personnelUpdate(projectId: string, contractId: string, personnelId: string): string {
```
MartinForre: Resource collection should not care about the "action" you're performing on a resource. This method could be merged with `personnel` with optional `personnelId` parameter. E.g. `personnel(projectId: string, contractId: string, personnelId?: string) { return combineUrls(this.contract(projectId, contractId), 'resources', 'personnel', personnelId);`

PR #33, opened 2020-03-13
Api - Added exception middleware with telemetry tracing

HansDahle:


src/backend/api/Fusion.Resources.Api/Middleware/ApiProblem.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System.Collections.Generic;
+using System.Net;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Fusion.Resources.Api.Middleware
+{
+    public class ApiProblem
```
leifbjarte: Dotnet core has a built-in class for this: 
https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.problemdetails?view=aspnetcore-3.1 
which adheres to IETF specification.

Can use helper method Problem(title: "Issues"....) to return errors in this format.

Perhaps using the standard dotnet core way of doing this will save us some work?

src/backend/api/Fusion.Resources.Api/Startup.cs
```diff
@@ -104,6 +113,32 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             app.UseEndpoints(endpoints =>
             {
                 endpoints.MapControllers();
+
+                // TODO: Remove
```
leifbjarte: Remove?

HansDahle: The whole point is to create a layer that will produce errors so martin can test errorhandling in the api. However we do not intend to let this make it's way to production, after we are done testing this.. Therefore comments for removal..

PR #43, opened 2020-03-19
Api - Support for contractor change requests

HansDahle:
Change requests will update an existing position. 

To create the change requests the POST must include `originalPositionId` that points to an existing org chart position. 

The api will validate that the position actually exists when verifying the input, and validate that the position belongs to the correct contract and project when executing the command.

The api will resolve the position targeted when returning the request object. New properties are 
```
{
  "originalPositionId": "nullable guid",
  "originalPosition": { "same as position property" },
  "originalPerson": { "api person" }
}
```

When fetching all requests, the `originalPosition` and `originalPerson` properties are null. This is due to the fact we need to resolve these from the org chart to get the info. To include these properties one can call the endpoint with `?$expand=originalPosition`.


HansDahle:
Response when context is incorrect for originalPosition. 
These occur if the position exist but ex. is in the orgchart and not the contract.

```json
{
    "title": "Invalid Operation",
    "status": 400,
    "detail": "Position does not belong to the correct contract. It belongs to Contract 002 (0000000002)",
    "error": {
        "code": "InvalidOrgChartPositionError",
        "message": "Position does not belong to the correct contract. It belongs to Contract 002 (0000000002)"
    }
}
```

HansDahle:
Validation error on request

Note: according to problem specification https://tools.ietf.org/html/rfc7231.
The title will be consisten for the same types of error, so this text can be used to identify input validation errors...

```json
{
    "type": "https://tools.ietf.org/html/rfc7231#section-6.5.1",
    "title": "One or more validation errors occurred.",
    "status": 400,
    "traceId": "00-8da8a925a7dfa94f942998c981a9c01f-3aa0be19de34984a-00",
    "errors": {
        "originalPositionId": [
            "Position with id 'bcc05871-dc9f-4939-be17-f71635c211ca' does not exist in org service."
        ]
    }
}
```

src/backend/api/Fusion.Resources.Api/Controllers/Requests/RequestsController.cs
```diff
@@ -20,38 +23,72 @@ namespace Fusion.Resources.Api.Controllers
...
+                    return new ApiContractPersonnelRequest(query);
+                }
+            }
+            catch (InvalidOrgChartPositionError ex)
+            {
+                return ApiErrors.InvalidOperation(ex);
+            }
+            catch (InvalidOperationException ex)
+            {
+                return ApiErrors.InvalidOperation(ex);
```
leifbjarte: Nice!

PR #50, opened 2020-03-23
Person Details overview and edit 

olerichard:
each person in Manage personnel overview is now clickable. 
This open a new sidesheet with person details. 
Has a general pane, with personnal info. 
And a positions pane showing active and past posistion. 

General pane has edit button to edit personnel info. 
can change name,phone and diciplines.


src/frontend/src/models/Personnel.ts
```diff
@@ -2,6 +2,88 @@ export type PersonnelDiscipline = {
...
+export type Requests = {
+    id: string;
+    created: Date;
+    updated: Date;
+    state: number;
+    position: RequestPosition;
+    contract: RequestContract;
+    project: RequestProject;
+};
+
```
eskil0312: Don't we have almost every ones of these models already defined? Look at `PersonnelRequest.ts`

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/PersonnelInfoSideSheet/PositionsTab/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+        workload: p.workload,
+    };
+};
+
+const PositionsTab: React.FC<PositionsTabProps> = ({ person }) => {
+
+    const positions = React.useMemo(() => person.positions?.map(mapPositionsToPersoPosition), [person]);
+
+    const currentDate = React.useMemo(() => new Date().getTime(), [])
+    const pastPositions = React.useMemo(() => positions?.filter(p => (p.appliesTo?.getTime() || currentDate) <= currentDate), [positions]);
```
eskil0312: Seems like this needs some Prettier magic?

src/frontend/src/models/Personnel.ts
olerichard: Yes and no. 
They exist but there is always some discrepancy.
which could easily make it a mess if I tried to "shoehorn" those models in. 
 
Decided to just write out the entire model I found in swagger for the personnel endpoint. 
And put it into this one model.

eskil0312: If the backend returns different models than `PersonnelRequest.ts` is mapped towards, I guess it is okay

PR #56, opened 2020-03-24
Event-based profile sync

leifbjarte:


leifbjarte:
Will reopen once the base branch is merged.

leifbjarte:
We might benefit from caching the personnel in the .../refresh endpoint to avoid spamming the DB, ref. issue with People. Thoughts?

src/backend/api/Fusion.Resources.Domain/Services/ProfileServices.cs
```diff
@@ -35,20 +35,25 @@ public ProfileServices(IFusionProfileResolver profileResolver, ResourcesDbContex
             return existingEntry;
         }
 
-        public async Task<DbExternalPersonnelPerson?> RefreshExternalPersonnelAsync(PersonId personId)
+        public async Task<DbExternalPersonnelPerson> RefreshExternalPersonnelAsync(PersonId personId)
         {
             var profile = await ResolveProfileAsync(personId);
             var resolvedPerson = await ResolveExternalPersonnelAsync(personId);
 
             //in the unlikely event of non-existent person, add it to database
```
HansDahle: Did not catch this the last round, but, in what scenario are we refreshing a profile that is not in the database? Should we not just throw on that? If it's not in the database, there is no need to get any latest changes...?

src/backend/function/Fusion.Resources.Functions/Functions/ProfileSync.cs
```diff
@@ -22,6 +24,24 @@ public ProfileSync(IHttpClientFactory httpClientFactory)
...
+        [FunctionName("profile-sync-event")]
+        public async Task SyncProfile(
+            [EventSubscriptionTrigger(HttpClientNames.Application.People, "subscriptions/persons", "resources-profile")] MessageContext message,
+            ILogger log)
+        {
+            log.LogInformation("Profile sync event received");
+            log.LogInformation(message.Event.Data);
+            var body = message.GetBody<PeopleSubscriptionEvent>();
+
+            if (body.Type != PeopleSubscriptionEventType.ProfileUpdated) //manual filtering to avoid hard-coding in attribute
```
HansDahle: The property checked here is not available for filtering when setting up the event subscription. So the comment can be removed as misleading?

src/backend/function/Fusion.Resources.Functions/Functions/ProfileSync.cs
```diff
@@ -22,6 +24,24 @@ public ProfileSync(IHttpClientFactory httpClientFactory)
...
+                return;
+
+            var refreshResponse = await resourcesClient.PostAsJsonAsync($"resources/personnel/{body.Person.Mail}/refresh", new { });
```
HansDahle: Should we maybe do a get first, to check if we even track the profile?

leifbjarte: The check is done in the refresh endpoint anyway. Given the amount of changes that "strangled" People i was thinking that it's better to reduce traffic to API?

src/backend/function/Fusion.Resources.Functions/Functions/ProfileSync.cs
leifbjarte: Ok, my misunderstanding, will remove.
I am quite confused as to what is the difference between "PeopleEventTypes" and "PeopleSubscriptionEventType". Perhaps other users might be as well, and some docs/comments on this is in place?

src/backend/api/Fusion.Resources.Domain/Services/ProfileServices.cs
leifbjarte: Good point, will change this.

src/backend/function/Fusion.Resources.Functions/Functions/ProfileSync.cs
HansDahle: The difference is that one type is on the subscription, the other is to fine grain the event.. To use filters on the service bus subscription, the property needs to be promoted to the service bus message headers. To do that it needs to be generalized in some sort --> SubscriptionEventType..  

But inside the message there might be more details, to not have too many changes in the subscription event constants, but one might still want to determine if the event should be acted upon etc.. 

Subscription type: `Profile`, `Companies`
PeopleEventType: `ProfileUpdated`, `ProfileDeleted`, `GPRDClearoutsale`, `AccountSignin`
CompanyEventType: `CompanyUpdated`, `CompanyCreated`, `CompanyDeprecated` etc... 

If the consuming service have no use for companies, the subscription can be filtered so no cpu is spent on useless messages.

src/backend/function/Fusion.Resources.Functions/Functions/ProfileSync.cs
HansDahle: The ppl situation is a bit different, because that gets the whole dump from AD.. 

However the ppl service only sends out events for profiles that have actual changes we care about (not all ad properties that we care about, like licenses etc), and if the profile is tracked in the ppl database. This will not result in the same flooding.

HansDahle: I guess one reason to maybe do a get first, is to reduce the amount of "legit" errors in AI logs? This would light up the 404 requests.. Dunno tbh.. But since the endpoint checks it should be good like this.

src/backend/function/Fusion.Resources.Functions/Functions/ProfileSync.cs
```diff
@@ -22,6 +24,31 @@ public ProfileSync(IHttpClientFactory httpClientFactory)
...
+
+            if (refreshResponse.StatusCode == System.Net.HttpStatusCode.NotFound)
+            {
+                log.LogWarning($"Person with email '{body.Person.Mail}' not found in Resources");
+                return;
+            }
+
+            refreshResponse.EnsureSuccessStatusCode();
```
HansDahle: This will not give us any info in the logs on what went wrong. A better way would be to log the content if the status code is nonsuccessfull...
```
if (!refreshResponse.IsSucce...) {
   var content = await resp.Content.ReadAsStringAsync();.
   log.LogError(content);
   resp.EnsureResp...();
}

File: src/backend/function/Fusion.Resources.Functions/Functions/ProfileSync.cs (hunk 3 above)
Comment (leifbjarte) [reply src/backend/function/Fusion.Resources.Functions/Functions/ProfileSync.cs]: Ok, leaving it like this for now.

File: src/backend/api/Fusion.Resources.Domain/Services/ProfileServices.cs [hunk 5]
```diff
@@ -60,7 +65,7 @@ public ProfileServices(IFusionProfileResolver profileResolver, ResourcesDbContex
                 resolvedPerson.AzureUniqueId = profile.AzureUniqueId;
                 resolvedPerson.JobTitle = profile.JobTitle;
                 resolvedPerson.Name = profile.Name;
-                resolvedPerson.Phone = profile.MobilePhone;
+                resolvedPerson.Phone = profile.MobilePhone ?? string.Empty; //column does not allow nulls, set empty string.
```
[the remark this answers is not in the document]
leifbjarte: Because reasons. Discussed and resolved this with Hans, can include you when this comes up again.

PR #71, opened 2020-03-26
Implemented un-allocated contracts

leifbjarte:


src/backend/api/Fusion.Resources.Api/Controllers/Contracts/ContractsController.cs
```diff
@@ -55,23 +52,41 @@ public async Task<ActionResult<ApiContract>> GetProjectContract([FromRoute]Proje
...
-            };
-
-            return Ok(new ApiCollection<ApiUnallocatedContract>(contracts));
+            if (context == null)
+                return ApiErrors.NotFound($"/contexts/project/{projectIdentifier.ProjectId}");
+
+            var projectMasterContext = await contextResolver.RelationsFirstOrDefaultAsync(context, FusionContextType.ProjectMaster);
+
+            if (projectMasterContext == null)
+                return ApiErrors.NotFound($"/contexts/projectMaster");
```
HansDahle: Should we implement this as an extension method in the integration lib? 
Think getting from a org chart to project master would be a common operation.. Could also utilize the path query endpoint..

Could maybe for now create an extension class in the resources project, 

```
public static class ContextResolverExtensions 
{
    public static Task<FusionContext> ResolveProjectMasterAsync(this IFusionContextResolver resolver, ProjectIdentifier projectid) 
    { /* The implementation highligted */ }
}
```

Would improve readbility.

src/backend/api/Fusion.Resources.Api/Fusion.Resources.Api.csproj
```diff
@@ -14,9 +14,9 @@
     <PackageReference Include="FluentValidation.AspNetCore" Version="8.6.2" />
     
     <PackageReference Include="Bogus" Version="29.0.1" />
-    <PackageReference Include="Fusion.Integration" Version="3.1.0-preview-7" />
-    <PackageReference Include="Fusion.Integration.Authorization" Version="3.1.0-preview-7" />
-    <PackageReference Include="Fusion.Integration.Org" Version="3.1.0-preview-7" />
+    <PackageReference Include="Fusion.Integration" Version="3.1.0-local-2" />
```
HansDahle: I have reserved local-1 through 4 on my computer ;)

leifbjarte: Ok, will do local-100 as my starting point, just in case :D

src/backend/api/Fusion.Resources.Api/Controllers/Contracts/ContractsController.cs
leifbjarte: Good suggestion, will do it.

src/backend/api/Fusion.Resources.Api/Controllers/ApiErrors.cs
```diff
@@ -42,12 +42,12 @@ internal static ActionResult InvalidPageSize(string message)
             };
         }
 
-        internal static ActionResult NotFound(string resourcePath)
```
HansDahle: The rfc allows for an instance to be defined. Perhaps allow for bot message and optional resource path?

src/backend/api/Fusion.Resources.Api/Controllers/Extensions/ContextResolverExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+            if (projectMasterContext == null)
+                throw new ProjectMasterNotFoundError(identifier);
+
+            return projectMasterContext;
+        }
+
+        public class ProjectMasterNotFoundError : Exception
+        {
+            public ProjectMasterNotFoundError(Resources.Api.Controllers.ProjectIdentifier projectIdentifier) :
+                base($"Unable to find master project for '{projectIdentifier.Name} ({projectIdentifier.OriginalIdentifier})'")
```
HansDahle: Should be project master, as that's the term being user.

PR #84, opened 2020-04-01
Bugfix - Task owner instance not rendering

eskil0312:
### Issue:
When dealing with org positions one have to deal with time. Right now position instances is filtered on today's date. When a position is expired(future/past) the instance will not show.
### Solution:
Have not decided witch solution to go for, but I suggest one of the following:
1. **Introduce future/past properties on position cards**: Show cards the way they are shown in org chart with a background color that indicate if the position is future or past
2. **Always select the first instance**. Most of the positions have only one instance. By just selecting the first instance no matter what we don't have to introduce future/past positions to the user.

 @MartinForre @HansDahle @olerichard


eskil0312:
After some consideration I'm voting for the first option. It's the most correct solution and it could possibly save us some work in the future if positions with more then one instance is a more common scenario. We should then be clear in the info page about the future/past color code and what they mean

src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ContractDetailsPage/index.tsx
```diff
@@ -60,26 +61,30 @@ const PositionCardSkeleton = () => (
...
-const renderPosition = (position: Position | null) =>
-    position ? (
+const renderPosition = (position: Position | null) => {
+    if (!position) {
+        return 'N/A';
+    }
+    const now = new Date();
+    const instance = getInstances(position, now)[0];
+    const isFuture = isInstanceFuture(instance, now);
+    const isPast = isInstancePast(instance, now);
```
MartinForre, line 72: useMemo on all of these?

HansDahle:
#1 would make sense yes..

PR #88, opened 2020-04-01
Manage personnel fetch issues

olerichard:
Manage Personnel fetched personnel twice.
Causing some issues where positions some times was not fetched properly.
Flickering data workload and positions data. 

Removed getPersonnelWithPositionAsync, instead using the apiCall directly in fetchPersonnelAsnyc.
Replacing the original getPersonnelAsync call.


MartinForre:
@olerichard The reason personnel is fetched twice is because the second one (with positions) will, with larger amount of personnel than our test contracts, take quite a bit longer since the API has to fetch positions for each person. That's why we fetch all personnel without positions first (which is quick) and display those while performing the slower fetch with positions in the background

PR #95, opened 2020-04-02
Error handling update for Requests.

olerichard:
new model RequestValidatioError. a combination of standard httpErrorResponse and response from backend validation.
Model implemented as error type for RequestProgressSideSheet.
RequestSidesheet wil now display errors as before but also from the new errors object.
errors object holds key for each field that has an error.
Each field can have multiple errors.
errors object get reduced down to an array of string and displayed for each invalid request.


src/frontend/src/components/RequestProgressSidesheet/index.tsx
```diff
@@ -75,6 +75,14 @@ function InvalidRequestProgressItem<TRequest>({
     onRemove,
 }: FailedRequestItemProps<TRequest>) {
     const ignoreTooltipRef = useTooltipRef('Ignore');
+
+    let errors: string[] = [];
```
eskil0312, line 79: Is this the same as `const errors = error.errors ? Object.values(error.errors) : []`;?

MartinForre: Looks like the error.errors[index] is an array? So more like `const errors = Object.values(error.errors || []).reduce((all, e) => ([ ...all, ...e ]), []);` to flatten the array of arrays?

PR #101, opened 2020-04-06
Api - Endpoint for parsing spreadsheet

HansDahle:
POST /utilities/parse-spreadsheet

--form 'file=@/C:/.../test-person-list.xlsx'

```
{
    "headers": [
        {
            "cellRef": "A1",
            "title": "Last Name",
            "colIndex": 0
        },
        {
            "cellRef": "B1",
            "title": "First Name",
            "colIndex": 1
        },
        {
            "cellRef": "C1",
            "title": "Mail",
            "colIndex": 2
        },
        {
            "cellRef": "D1",
            "title": "Phone",
            "colIndex": 3
        }
    ],
    "data": [
        {
            "index": 0,
            "items": [
                "[Last name]",
                "[First name]",
                "mail@company.com",
                "[phone]"
            ]
        },
	...
 
    ],
    "messages": []
}
```

### DIFF
--- a/src/backend/api/Fusion.Resources.Api/Controllers/Utilities/UtilitiesController.cs
+++ b/src/backend/api/Fusion.Resources.Api/Controllers/Utilities/UtilitiesController.cs
@@ -1,0 +1,140 @@
+﻿using Fusion.Integration.Configuration;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading.Tasks;
+
+namespace Fusion.Resources.Api.Controllers.Utilities
+{
+    [Authorize]
+    [ApiController]
+    public class UtilitiesController : ResourceControllerBase
+    {
+        private readonly IHttpClientFactory httpClientFactory;
+        private readonly IFusionTokenProvider tokenProvider;
+        private readonly IConfiguration configuration;
+        private readonly IOptions<FusionIntegrationOptions> fusionOptions;
+
+        public UtilitiesController(IHttpClientFactory httpClientFactory, IFusionTokenProvider tokenProvider, IConfiguration configuration, IOptions<FusionIntegrationOptions> fusionOptions)
+        {
+            this.httpClientFactory = httpClientFactory;
+            this.tokenProvider = tokenProvider;
+            this.configuration = configuration;
+            this.fusionOptions = fusionOptions;
+        }
+
+        [HttpPost("/utilities/parse-spreadsheet")]
+        public async Task<ActionResult<ExcelConversion>> ValidateContractorImportSpreadsheet([FromForm]ConvertSpreadsheetRequest request)
+        {
+            if (request == null)
+                return FusionApiError.InvalidOperation("MissingBody", "Could not locate any body payload");
+
+            var url = $"https://pro-f-utility-{fusionOptions.Value.Environment}.azurewebsites.net";
+            var token = await tokenProvider.GetApplicationTokenAsync();
+            var client = httpClientFactory.CreateClient();
+            client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("bearer", token);
+            client.BaseAddress = new Uri(url);
+
+
+            if (string.IsNullOrEmpty(url))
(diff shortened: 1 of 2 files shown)


HansDahle:
No file provided

```json
{
    "type": "https://tools.ietf.org/html/rfc7231#section-6.5.1",
    "title": "One or more validation errors occurred.",
    "status": 400,
    "traceId": "00-f4b5b45918c0a74f95b4850578b28f36-002295154963c040-00",
    "errors": {
        "File": [
            "File must be provided"
        ]
    }
}
```

HansDahle:
Invalid file

```
{
  "title": "Unhandle error",
  "type": "about:blank",
  "status": 500,
  "details": "Parser function returned non-successfull response (BadRequest).",
  "instance": null,
  "error": {
    "code": "InternalServerError",
    "message": "Parser function returned non-successfull response (BadRequest)."
  }
}
```

HansDahle:
Looks like 404 is returned when the file is too large... With html body :/ This is done in the web.conf request filter. Will probably be something else in another ingress..

MartinForre:
It would be better, and easier, for the frontend if the error response was in a "standardized" format

HansDahle:
They are, using the problem RFC specification.

The { error: { } } is added to support backwards compatibility. 
What do you refer to as not-standardized?

MartinForre:
error vs errors for instance. One has a traceId and the other has an instance.

HansDahle:
Errors are the standard way for .net core 3.1 to return model validation error. 
So the combination of title = "One or more validation errors occurred." and 400, the errors will contain all validation errors.

To override that we need to create overrides some built in stuff. Not gonne do that, at least in this branch...

src/backend/api/Fusion.Resources.Api/Controllers/Utilities/UtilitiesController.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            var url = $"https://pro-f-utility-{fusionOptions.Value.Environment}.azurewebsites.net";
+            var token = await tokenProvider.GetApplicationTokenAsync();
+            var client = httpClientFactory.CreateClient();
+            client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("bearer", token);
+            client.BaseAddress = new Uri(url);
+
+
+            if (string.IsNullOrEmpty(url))
+                throw new InvalidOperationException("Missing configuration for fusion utility function");
```
leifbjarte, line 48: Is there a specific reason why is this not also using FusionApiError.InvalidOperation(...) rather than throwing exception directly?

src/backend/api/Fusion.Resources.Api/Controllers/Utilities/UtilitiesController.cs
```diff
@@ -0,0 +1,140 @@
...
+                streamContent.Headers.ContentLength = request.File.Length;
+
+                var response = await client.PostAsync("/api/excel-json-converter", streamContent);
+
+                var content = await response.Content.ReadAsStringAsync();
+
+                if (response.IsSuccessStatusCode)
+                    return JsonConvert.DeserializeObject<ExcelConversion>(content);
+
+                throw new InvalidOperationException($"Parser function returned non-successfull response ({response.StatusCode}).");
```
leifbjarte, line 63: Same goes here, I guess.

src/backend/api/Fusion.Resources.Api/Controllers/Utilities/UtilitiesController.cs
HansDahle: Because it is not a user error, it is a configuration error, 500 not 400.

src/backend/api/Fusion.Resources.Api/Controllers/Utilities/UtilitiesController.cs
HansDahle: Same response i guess :)
I guess this could be a bad request, as the spread sheet could be invalid etc, but the effort to analyse and respond differently does not really outweigh the benefit.

src/backend/api/Fusion.Resources.Api/Controllers/Utilities/UtilitiesController.cs
leifbjarte: Ok - agree in regards to status code. We could still do FusionApiError.Internal(...)  with the same API model as FusionApiError.InvalidOperation(), but it's not a big deal.

src/backend/api/Fusion.Resources.Api/Controllers/Utilities/UtilitiesController.cs
leifbjarte: Agreed. See discussion in previous comment.

PR #117, opened 2020-04-07
Added API endpoints for request comments

leifbjarte:


src/backend/api/Fusion.Resources.Api/Controllers/Requests/ApiModels/ApiRequestComment.cs
```diff
@@ -1,21 +1,34 @@
...
+        public ApiRequestComment(QueryRequestComment queryComment)
+        {
+            Created = queryComment.Created;
+            CreatedBy = new ApiPerson(queryComment.CreatedBy);
+            Updated = queryComment.Updated;
+
+            UpdatedBy = queryComment.UpdatedBy != null ? new ApiPerson(queryComment.UpdatedBy) : null;
+            Content = queryComment.Content;
+
+            if (Enum.TryParse<ApiCommentOrigin>(queryComment.Origin, true, out var result))
```
HansDahle, line 18: Should also have a "Unknown" value in the enum to support when the parsing fails. Otherwise the "Company" will be used as that is index 0 and enum is not nullable.

src/backend/api/Fusion.Resources.Api/Controllers/Requests/Requests/CreateRequestComment.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Text.Json.Serialization;
+
+namespace Fusion.Resources.Api.Controllers
+{
+    public class CreateRequestComment
+    {
+        public string Content { get; set; } = null!;
+
+        [JsonConverter(typeof(JsonStringEnumConverter))]
+        public CommentOrigin Origin { get; set; }
```
HansDahle, line 11: Should this not be resolved by the api? If User.AccountType == employee -> Company, else contractor?

src/backend/api/Fusion.Resources.Api/Controllers/Requests/RequestsController.cs
```diff
@@ -321,8 +312,91 @@ public async Task<ActionResult<ApiContractPersonnelRequest>> ProvisionContractor
...
+            var authResult = await Request.RequireAuthorizationAsync(r =>
+            {
+                r.AlwaysAccessWhen().FullControl();
+
+                r.AnyOf(or =>
+                {
+                    or.ContractAccess(ContractRole.Any, projectIdentifier, contractIdentifier);
+                    or.BeContractorInContract(contractIdentifier);
+                });
+            });
```
HansDahle, line 358: Should everyone be able to update everyone's comments?

Should it be:
   AlwaysWhen -> FullControl
   r.Must().BeCommentAuthor(Guid commentId)

?

src/backend/api/Fusion.Resources.Database/Entities/DbRequestComment.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Fusion.Resources.Database.Entities
+{
+    public class DbRequestComment
+    {
+        public Guid Id { get; set; }
+
+        public string Comment { get; set; } = null!;
+
+        public string Origin { get; set; } = null!;
```
HansDahle, line 12: Use enum?

src/backend/api/Fusion.Resources.Domain/Commands/Requests/AddComment.cs
```diff
@@ -0,0 +1,59 @@
...
+            Origin = origin;
+        }
+
+        public Guid RequestId { get; }
+
+        public string Comment { get; }
+
+        public string Origin { get; }
+
+        public Guid CreatedById { get; set; }
```
HansDahle, line 26: Is this needed when using trackable request and editor?

src/backend/api/Fusion.Resources.Domain/Commands/Requests/AddComment.cs
```diff
@@ -0,0 +1,59 @@
...
+                var contractorRequest = await db.ContractorRequests.FirstOrDefaultAsync(c => c.Id == command.RequestId);
+
+                if (contractorRequest == null)
+                    throw new InvalidOperationException($"Contractor request with id '{command.RequestId}' was not found");
+
+                var comment = new DbRequestComment
+                {
+                    Id = Guid.NewGuid(),
+                    Comment = command.Comment,
+                    Origin = command.Origin,
```
HansDahle, line 48: I guess the origin could be calculated here, using the editor dbperson?

src/backend/api/Fusion.Resources.Domain/Queries/GetContractPersonnelRequest.cs
```diff
@@ -53,14 +60,20 @@ public async Task<QueryPersonnelRequest> Handle(GetContractPersonnelRequest requ
...
-               
+
                 var position = new QueryPositionRequest(dbRequest.Position)
                     .WithResolvedBasePosition(basePosition);
 
                 var workflow = await mediator.Send(new GetRequestWorkflow(request.RequestId));
 
                 var returnItem = new QueryPersonnelRequest(dbRequest, position, workflow);
 
+                if (request.Query?.ShoudExpand("comments") ?? false)
```
HansDahle, line 71: I've tried to make the expands a part of the query request, in an attempt to make it available in pragmatically executed queries.. Maybe check if that could be done here as well?

src/backend/api/Fusion.Resources.Api/Controllers/Requests/Requests/CreateRequestComment.cs
leifbjarte: Of course! 🤦‍♂

src/backend/api/Fusion.Resources.Api/Controllers/Requests/RequestsController.cs
leifbjarte: Good point, will add auth req and handler for it.

src/backend/api/Fusion.Resources.Domain/Commands/Requests/AddComment.cs
leifbjarte: No. Started out this way and ended up using Editor, removing.

src/backend/api/Fusion.Resources.Domain/Commands/Requests/AddComment.cs
leifbjarte: Yes, good idea.

src/backend/api/Fusion.Resources.Domain/Queries/GetContractPersonnelRequest.cs
leifbjarte: Sorry, not quite sure what you mean here. Will talk about this one directly.

HansDahle: Convert as much of the odataqueryparam in the query object as possible. This makes the functionality available programmatically (typo above :p). I tried a pattern that defines a private enum (flag-enabled) and private Expand property on the command.

That way the handler can reach it, and checking for expand is more robust/clean as you don't have to do all the string stuff..

One can also when required create public fluent methods on the query to set the expand ( .ExpandComments() ) in case one needs it in a internal query...

HansDahle: Ref `GetContractPersonnelRequests`

leifbjarte: Implemented in similar pattern as `GetContractPersonellRequests`

src/backend/api/Fusion.Resources.Api/Controllers/Requests/RequestsController.cs
```diff
@@ -321,8 +312,103 @@ public async Task<ActionResult<ApiContractPersonnelRequest>> ProvisionContractor
...
+
+            if (comment == null)
+                return FusionApiError.NotFound(commentId, "Comment not found");
+
+            #region Authorization
+
+            var authResult = await Request.RequireAuthorizationAsync(r =>
+            {
+                r.AlwaysAccessWhen().FullControl();
+                r.AnyOf(r => r.BeCommentAuthor(comment));
```
HansDahle, line 357: This would cause that the user must be the comment author AND one of the requirements below. They should all be in the same block.

src/backend/api/Fusion.Resources.Api/Controllers/Requests/RequestsController.cs
```diff
@@ -321,8 +312,103 @@ public async Task<ActionResult<ApiContractPersonnelRequest>> ProvisionContractor
...
```
HansDahle, line 389: Move to the block below to have a proper OR block.

src/backend/api/Fusion.Resources.Domain/Queries/GetContractPersonnelRequest.cs
```diff
@@ -1,27 +1,56 @@
...
+        private ODataQueryParams query = null!;
+
         public GetContractPersonnelRequest(Guid requestId)
         {
             RequestId = requestId;
         }
 
         public Guid RequestId { get; }
 
+        public ODataQueryParams Query
```
HansDahle, line 24: Should we make the setter private, to force the use of the method. Then we can contain all the logic regarding odata query in the method, not spread around...

src/backend/api/Fusion.Resources.Api/Controllers/Requests/RequestsController.cs
leifbjarte: If I put them in the same block, would that not allow users to edit other users' comments?

src/backend/api/Fusion.Resources.Api/Authorization/Requirements/RequestCommentAuthor.cs
```diff
@@ -0,0 +1,28 @@
...
+        public Task HandleAsync(AuthorizationHandlerContext context)
+        {
+            var azureId = context.User.GetAzureUniqueId();
+
+            var resource = context.Resource as QueryRequestComment;
+
+            if (resource?.CreatedBy.AzureUniqueId == azureId)
+                context.Succeed(this);
+
+            SetEvaluation("User is not author of the comment");
```
HansDahle, line 23: Could maybe detect that the resource is not the expected type, and report that the requirement was not evaluated. 
```
Evaluation = "Invalid comment resource passed in"
WasEvaluated = false
```
?

PR #119, opened 2020-04-07
Added fusion-libraries to nuget config

leifbjarte:
Resources Functions uses Fusion.Events.Azure.Functions.Extensions which is published to the Fusion-Libraries feed. The build is failing as of now because of missing package.


leifbjarte:
> Nah, i don't want to use the protected feed. Too many issues to get builds going etc. I've added to most of the library builds to publish to both feeds. So we need to make sure that package is published to the public feed instead of going back. Gonne cause issue on the docker build etc...

Ok. Will abandon this PR then.

PR #130, opened 2020-04-15
10974 Manage personnel without access - a.filter is not a function

olerichard:


src/frontend/src/pages/ProjectPage/pages/ContractPage/pages/ManagePersonnelPage/index.tsx
```diff
@@ -35,16 +35,14 @@ const ManagePersonnelPage: React.FC = () => {
...
-            return error;
-        }
+        const result = await apiClient.getPersonnelWithPositionsAsync(projectId, contractId);
+        dispatchContractAction({
+            verb: 'merge',
+            collection: 'personnel',
+            payload: result,
+        });
+
+        return error;
```
eskil0312, line 45: Where is error defined?

olerichard: was part of the try catch, Ive removed it now.

PR #146, opened 2020-06-24
Bug/excel import - Batched requests closes sidesheet

olerichard:
When using savePersonnelCollectionAsync setFailedRequest was set regardless off any personnel failed to be saved.

The update off setFailedRequests triggered a useEffect inside RequestProgressSidesheet that would close the sidesheet if failedRequest was 0.
In most cases this would cause the sidesheet to close when the first collection was finnished regardless off pending requests.

This has been fixed by setting the response.reduce that finds BadRequests to a variable and only setting it to setFailedRequests if it contains something.


src/frontend/src/components/RequestProgressSidesheet/index.tsx
```diff
@@ -186,7 +186,17 @@ function RequestProgressSidesheet<TRequest, TResponse>({
...
 
     return (
-        <ModalSideSheet header={title} show={isShowing} onClose={onClose}>
+        <ModalSideSheet
+            header={title}
+            show={isShowing}
+            onClose={onClose}
+            safeClose={pendingRequests.length > 0}
+            safeCloseTitle={`
+                Closing this sidesheet wont stop the saving routine, but you will not be able to verify the result.
```
MartinForre, line 195: `won't`

PR #167, opened 2020-10-07
Delete position & integration test pilot

HansDahle:
- [x] New feature
- [x] Bug fix
- [ ] High impact

**Description of work:**
Fixed contract list endpoint for project, which did not display contracts where the user is only delegated admin (without having a position).

Added new endpoints to facilitate deleting positions. This is done to prevent users having edit permissions directly in the org chart, which would give them the rights to add whomever they want without going through the requests.

Added options endpoint to check if the user have access to delete positions inside a contract.

Started adding a pilot for integration testing support for integration lib. These projects will be moved to packages when they support a majority of the feature and in a usefull syntax.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

USing postman, add a position in a contract and remove it using /projects/{orgProjectId}/contracts/{orgContractId}/mpp/positions/{orgPositionId}

**Checklist:**
- [x] Considered security

### DIFF
--- a/src/backend/Directory.Build.props
+++ b/src/backend/Directory.Build.props
@@ -1,0 +1,12 @@
+<Project>
+  <Target Name="_ResolveCopyLocalNuGetPackagePdbsAndXml" Condition="$(CopyLocalLockFileAssemblies) == true" AfterTargets="ResolveReferences">
+    <ItemGroup>
+      <ReferenceCopyLocalPaths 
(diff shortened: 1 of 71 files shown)


HansDahle:
Don't to reviews on the testing parts in this PR. We can change that at a later stage. The setup that is added her now works so we can start to add tests, but we should discuss how to set up test data for the mock services later.

leifbjarte (approved):

PR #182, opened 2020-12-03
API - Notifications - Fixed bugs

leifbjarte:
- [ ] New feature
- [x] Bug fix
- [ ] High impact

**Description of work:**
Corrected already sent check and ignored time on active instance comparison

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

Test in FQA.

**Checklist:**
- [x] Considered work items 
- [x] Considered security


src/backend/api/Fusion.Resources.Domain/Commands/Contracts/UpdateContractReps.cs
```diff
@@ -67,7 +67,7 @@ protected override async Task Handle(UpdateContractReps request, CancellationTok
 
                 if (request.ContractResponsiblePositionId.HasBeenSet)
                 {
-                    if (request.ContractResponsiblePositionId.Value.HasValue && contract.CompanyRep?.Id != request.ContractResponsiblePositionId.Value.Value)
+                    if (request.ContractResponsiblePositionId.Value.HasValue && contract.ContractRep?.Id != request.ContractResponsiblePositionId.Value.Value)
```
xssic, line 70: Assuming the previous CompanyRep was incorrect here.

src/backend/api/Fusion.Resources.Api/appsettings.json
```diff
@@ -22,8 +22,9 @@
   },
   "Fusion": {
     "ServiceDiscovery": {
-      //"ProOrganisation": "https://localhost:44311"
-      //"people": "https://localhost:44332"
+      "ProOrganisation": "https://pro-s-org-fqa.azurewebsites.net",
```
xssic, line 25: Any specific reason for using QA (or maybe CI) instead of local environment settings ?

src/backend/api/Fusion.Resources.Domain/Commands/Contracts/UpdateContractReps.cs
HansDahle: By the if check, the wrong property is used, yes

HansDahle (approved):

src/backend/api/Fusion.Resources.Domain/Extensions/PositionV2Extensions.cs
```diff
@@ -28,7 +28,7 @@ public static bool HasActiveAssignment(this ApiPositionV2? position, Guid azureU
             if (position.Instances == null)
                 return null;
 
-            return position.Instances.FirstOrDefault(i => i.AppliesFrom <= DateTime.Today && i.AppliesTo >= DateTime.Today);
+            return position.Instances.FirstOrDefault(i => i.AppliesFrom.Date <= DateTime.Today && i.AppliesTo.Date >= DateTime.Today);
         }
```
xssic, line 32: Looks OK.
When using Date component you would get from 00:00:00, so I guess only the first modification here will make any difference.

Maybe you could consider verifying/validating extension HasActiveAssignment (in same file), regarding AppliesFrom :-)
Last but not least. There is also a similar extension in the API project called HasActiveAssignment.

xssic (approved):

src/backend/api/Fusion.Resources.Api/appsettings.json
leifbjarte: Just for local testing. Commented these out, using default settings.

src/backend/api/Fusion.Resources.Domain/Commands/Contracts/UpdateContractReps.cs
leifbjarte: Indeed - that is the bugfix,

src/backend/api/Fusion.Resources.Domain/Extensions/PositionV2Extensions.cs
leifbjarte: Good catch, will consolidate these. The one in the API can actually be removed.

PR #209, opened 2021-02-15
add controller for sector/department information

loldot:


HansDahle:
Think we can close this?

PR #215, opened 2021-02-18
Api versioning and swagger improvements

HansDahle:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
Added api versioning support with default 1.0. This should let all existing endpoints work as execpted without having to provide a `?api-version=` param.

As uasability of swagger degrades when version is added, as the query param is not defined anywhere, we must add support for this by manipulating the output through swagger filters.
This filter adds a api-version query param to all endpoints and sets default value for a specific document. This means different versions must be added to different documents.

By making pages for different versions the usability of swagger ui improves, however using swagger to generate clients etc could see a 'hit'.

Also added swagger filters to make the complex models used in odata filtering usable, like removing the properties and replacing with a single `$filter` query string.

**Testing:**
- [x] Can be tested
- [ ] ~~Local tests are passing~~

Open swagger for pr api.

**Checklist:**
- [x] Considered security

### DIFF
--- a/src/backend/api/Fusion.Resources.Api/Startup.cs
+++ b/src/backend/api/Fusion.Resources.Api/Startup.cs
@@ -37,8 +37,18 @@ namespace Fusion.Resources.Api
                 });
 
 
+            services.AddApiVersioning(s =>
+            {
+                s.ReportApiVersions = true;
+                s.AssumeDefaultVersionWhenUnspecified = true;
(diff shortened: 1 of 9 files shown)


HansDahle:
Swagger should be available @ https://resources-api-pr-215.fusion-dev.net/swagger

xssic (approved):

PR #220, opened 2021-02-22
Get department requests endpoint

mhegg:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
Added GET endpoint for returning requests assigned to department 

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security


src/backend/tests/Fusion.Resources.Api.Tests/IntegrationTests/InternalResourceAllocationRequestTests.cs
```diff
@@ -275,7 +277,50 @@ public async Task PostRequest_InvalidRequest_ShouldBe_Unsuccessful()
...
+               .WithProject(testProject.Project)
+               .WithProposedPerson(testProfile)
+               .WithAssignedDepartment("Different department");
+            ;
+            await Client.TestClientPostAsync($"/projects/{anotherTestRequest.Project.ProjectId}/requests", anotherTestRequest.Request, new { Id = Guid.Empty });
+            var response = await Client.TestClientGetAsync<PagedCollection<ResourceAllocationRequestTestModel>>($"/departments/{anotherTestRequest.Request.AssignedDepartment}/resources/requests");
+            response.Should().BeSuccessfull();
+
+            response.Value.Value.Count().Should().Be(1);
+            AssertPropsAreEqual(response.Value.Value.First(), anotherTestRequest.Request, adminScope);
```
HansDahle, line 297: I guess this is a copy paste related, but should consider what we are testing in each test. Could have a general talk on how to write these tests, but in essence you want to test as little as possible in a single test. 

In this case for instance, we are not really testing that the properties are set, there are other tests for that. Here we only really care about the requests returned have the correct assigned department. 

Also the request url used is incorrect. It should be the one you have created... `/departments/{Different department}/requests`.

Asserting this we should only care that the department is correct.

```
response.Value.Value.Should().HaveCount(1);
response.Value.Value.Should().OnlyContain(r => r.AssignedDepartment = "Different department");
```

In the setup you might just want to create a single request, with that assigned department. 

Now you are sure that requests are returned from the endpoint..
Then you might want to verify negatives... `GetDepartmentRequests_ShouldNotIncludeRequests_WhenAssignedDepartmentEmpty`, `GetDepartmentRequests_ShouldNotIncludeRequests_WhenAssignedDepartmentNotCurrentDepartment` etc... 
In the setup here you might still just want to create a single request, with `assignedDepartment = null` or random string. 

The combination of these tests will ensure a correct behaviour.. If all tests test and assert the same things, you don't know what fails when they fail... Everything could go red....

HansDahle (commented):
Nice start

src/backend/tests/Fusion.Resources.Api.Tests/IntegrationTests/InternalResourceAllocationRequestTests.cs
mhegg: Not sure what you mean about the url being incorrect? The one I created is /departments/{Different department}/resources/requests, as described in the spec.

Or did you just mean that I should hard code the department string in the test? departments/Different department/resources/requests

HansDahle: My bad. didn't see the request under..

HansDahle (approved):

PR #229, opened 2021-03-02
Internal requests - Endpoint returning count of unassigned requests

xssic:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
Returns a count of unassigned requests.
Requests without assigned departments.

Can pass optional queryString argument isDraft to distinct between requests in draft or not.
Default counts all unassigned requests

/resources/requests/internal/unassigned?isdraft=false
/resources/requests/internal/unassigned?isdraft=true
/resources/requests/internal/unassigned

**Testing:**
- [ ] ~~Can be tested~~
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security

### DIFF
--- a/src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
+++ b/src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
@@ -225,6 +225,27 @@ namespace Fusion.Resources.Api.Controllers
             }
         }
 
+        [HttpGet("/resources/requests/internal/unassigned")]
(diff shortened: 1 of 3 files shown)


HansDahle:
This is not really how we do it :) 
These endpoints looks like they are returning data, but all they return is an int. This is not like any other pattern we do.

The count is a metadata, if we only want to return metadata we should do it in some other way, that makes sense. 
This resource path should be use to return the actual requests that are unassigned. We need this path, so we can allow more access to these requests. 

My suggestion would be to implement a HEAD call to this resource path, where we could implement the rang header values to keep the count.
This is a pattern that can be reused on other collections. 

Another solution could be to add support for a new `?$..` parameter, which only returned the count in the body. But i think this needs a bit more evaluation.

PR #235, opened 2021-03-08
move department/sector info to db

loldot:
Move department information to database.


src/backend/api/Fusion.Resources.Database/Entities/DbDepartment.cs
```diff
@@ -0,0 +1,102 @@
+﻿using Microsoft.EntityFrameworkCore;
+using System.Collections.Generic;
+
+namespace Fusion.Resources.Database.Entities
+{
+    public class DbDepartment
```
HansDahle, line 6: To resolve all sectors, we would just select distinct sector ids?

src/backend/api/Fusion.Resources.Domain/Queries/GetDepartments.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Fusion.Resources.Domain
+{
+    public class GetDepartments : IRequest<IEnumerable<QueryDepartment>>
+    {
+        public string? DepartmentFilter { get; set; }
+        public string Sector { get; set; }
```
HansDahle, line 16: I think i would make these properties private, and expose the usecase through builder methods, like mentioned earlier.. 

Just think of the next poor guy, to understand what these properties do... :)

HansDahle: Sector should also be nullable

src/backend/tests/Fusion.Resources.Api.Tests/Fixture/ResourceApiFixture.cs
```diff
@@ -60,11 +61,30 @@ public void Dispose()
...
-            var account = new FusionTestUserBuilder(accountType)                
+            var account = new FusionTestUserBuilder(accountType)
                 .SaveProfile();
 
-            
+
             return account;
         }
+
+        internal void RequireDepartment(string departmentId, string sectorId = null)
```
HansDahle, line 71: Require? Should this be ensure?

src/backend/api/Fusion.Resources.Database/Entities/DbDepartment.cs
loldot: The idea, for now at least, was that sectors would be where DepartmentId == SectorId, but distinct would also work

src/backend/api/Fusion.Resources.Api/Controllers/Person/PersonController.cs
```diff
@@ -28,11 +30,13 @@ public class PersonController : ResourceControllerBase
     {
         private readonly IHttpClientFactory httpClientFactory;
         private readonly IFusionProfileResolver profileResolver;
+        private readonly IMediator mediator;
 
-        public PersonController(IHttpClientFactory httpClientFactory, IFusionProfileResolver profileResolver)
+        public PersonController(IHttpClientFactory httpClientFactory, IFusionProfileResolver profileResolver, IMediator mediator)
```
HansDahle, line 35: This is handled in the base controller.. You can use it by just using the protected Dispatch method..

src/backend/api/Fusion.Resources.Database/Entities/DbDepartmentResponsible.cs
```diff
@@ -0,0 +1,52 @@
...
+    public class DbDepartmentResponsible
+    {
+        public Guid Id { get; set; }
+        public string DepartmentId { get; set; } = null!;
+        public Guid ResponsibleAzureObjectId { get; set; }
+        public DateTimeOffset DateFrom { get; set; }
+        public DateTimeOffset DateTo { get; set; }
+
+        public DateTimeOffset DateCreated { get; set; }
+        public DateTimeOffset DateUpdated { get; set; }
```
HansDahle, line 16: Updated should be nullable date

HansDahle (approved):

PR #245, opened 2021-03-17
Api - Allocation request type refactor

HansDahle:
- [ ] New feature
- [ ] Bug fix
- [x] High impact

**Description of work:**
Refactored request type to simply allocation for normal/direct/jv and moved to sub type. The subtype is resolved automatically if not provided. 

**Testing:**
- [ ] Can be tested
- [ ] Local tests are passing

TBD

**Checklist:**
- [ ] Considered work items 
- [ ] Considered security


src/backend/api/Fusion.Resources.Database/Migrations/20210317154326_Added subtype to request.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Fusion.Resources.Database.Migrations
+{
+    public partial class Addedsubtypetorequest : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "SubType",
+                table: "ResourceAllocationRequests",
+                type: "nvarchar(max)",
```
loldot, line 12: I guess this is nitpicking, and a problem many other places, but we should probably avoid using nvarchar(max) for short, known(ish) length strings as nvarchar(max) values are stored outside the row in the database and requires more reads for each row. This is especially the case if we want to index the field.

src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/Allocation/CanApproveStepHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+                //var request = await dbContext.ResourceAllocationRequests.FirstAsync(r => r.Id == notification.RequestId);
+
+                switch (notification.CurrentStepId)
+                {
+                    case "proposed":
+
+                        break;
+                }
+
+                return Task.CompletedTask;
```
loldot, line 35: Should this be implemented?

src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/Allocation/InitializedHandler.cs
```diff
@@ -28,24 +26,45 @@ public InitializedHandler(ResourcesDbContext dbContext, IMediator mediator)
...
 
-                    var workflow = new InternalRequestDirectWorkflowV1(initiatedBy);
                     dbContext.Workflows.Add(workflow.CreateDatabaseEntity(notification.RequestId, DbRequestType.InternalRequest));
 
                     request.LastActivity = DateTime.UtcNow;
-                    request.State.State = InternalRequestDirectWorkflowV1.CREATED;
+                    request.State.State = AllocationDirectWorkflowV1.CREATED;
+
+                    // Try to resolve the default assigned department
+                    request.AssignedDepartment = await mediator.Send(new Queries.ResolveResponsibleDepartment(request.Id));
```
loldot, line 59: Would this not overwrite the assigned department if it is sent through the API, or are we deprecating that option?

src/backend/api/Fusion.Resources.Database/Migrations/20210317154326_Added subtype to request.cs
HansDahle: Yeh, it's not even nitpicking.. forgot about this.

src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/Allocation/InitializedHandler.cs
HansDahle: It is sort of deprecated, but could add a nullcheck here, so we do not blindly overwrite..

src/backend/api/Fusion.Resources.Database/Migrations/20210317154326_Added subtype to request.cs
HansDahle: I think we could probably do that in a separate pull request, for many different entities

src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/Allocation/CanApproveStepHandler.cs
HansDahle: Added as placeholder, will implement when rules are more known.

loldot (approved):
💯

PR #252, opened 2021-03-30
Add endpoints for departments with responsible

loldot:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
Added endpoints get department information, including resource owner from line org, an de facto resource owner

Related docs PR: https://github.com/equinor/fusion/pull/100
**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security


HansDahle:
Hmm.. the search endpoint uses search to the line org, then for each result dispatches a query for each department, which in turn makes a request to line org.. There will be some scaling issues here..

It will be better to simply just query the line org for _all_ persons which are resource owners, then cache them for a while (this will not change often....).

Maybe instead of creating a query for this and a subquery with subqueries etc we should create a query that simply `GetDepartments`, then expose different expands on it, `.ExpandResponsible()` and `.WhereXYZ()`. This allows us to have one query that can operate efficiently. It can select multiple rows from the db, instead of one by one etc.

By caching the line org leaders, we will also insulate us a bit from transient errors. This could be cached ex. once a day etc.

We should try to reduce the amount of different query models `SomeEntityWithA`, then `SomeEntityWithSomethingElseAndTheAThing` etc.. Instead use nullable types and let them be null if not included?

loldot:
> 
> 
> Hmm.. the search endpoint uses search to the line org, then for each result dispatches a query for each department, which in turn makes a request to line org.. There will be some scaling issues here..

It only queries line org directly more than once when there are multiple pages. The other requests are to the people service, which I thought are cached?

> By caching the line org leaders, we will also insulate us a bit from transient errors. This could be cached ex. once a day etc.

True, but there are other trade-offs to consider when not operating on "live" data. Are we talking about caching in our database or setting up redis/something similar? The latter might be a bit much work for something that is likely to change in the near future.

> We should try to reduce the amount of different query models `SomeEntityWithA`, then `SomeEntityWithSomethingElseAndTheAThing` etc.. Instead use nullable types and let them be null if not included?

Agreed. I'll clean this up a bit

src/backend/api/Fusion.Resources.Domain/Queries/GetDepartments.cs
```diff
@@ -35,24 +49,102 @@ public GetDepartments StartsWith(string department)
...
+
+                    uri = page?.NextPage;
+                    var resourceOwners = page!.Value;
+
+                    foreach (var resourceOwner in resourceOwners)
+                    {
+                        if (!managedDepartments.ContainsKey(resourceOwner.FullDepartment)) continue;
+
+                        var department = managedDepartments[resourceOwner.FullDepartment];
+                        department.LineOrgResponsible = await profileResolver.ResolvePersonBasicProfileAsync(resourceOwner.AzureUniqueId);
```
HansDahle, line 133: Iterating over an unkown number of items, making 1 or more external api call is not scalable. The resolver cache the profile, but the initial hit can be problematic. These resolvers support batch processing for this reason. 

What i usually do is to select all the ids that are relevant to resolve, then resolve all of them in one go before the iteration of internal model.

src/backend/api/Fusion.Resources.Domain/Queries/GetDepartments.cs
```diff
@@ -35,24 +49,102 @@ public GetDepartments StartsWith(string department)
...
+            }
+
+            private async Task<List<QueryDepartment>> ExpandResourceOwners(List<QueryDepartment> departments, string? filter, CancellationToken cancellationToken)
+            {
+                var result = new List<QueryDepartment>();
+                var managedDepartments = departments.ToDictionary(dpt => dpt.DepartmentId);
+
+                var client = httpClientFactory.CreateClient("lineorg");
+
+                var uri = "/lineorg/persons?$filter=isresourceowner eq true";
```
HansDahle, line 110: There is no upper limit on the page size in the lineorg api. You could just set the page size to 2k. This is a plain lookup in the database for that api.

src/backend/api/Fusion.Resources.Domain/Models/LineOrg/ProfileWithDepartment.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Fusion.Resources.Domain.LineOrg
+{
+    public class ProfileWithDepartment
```
HansDahle, line 5: Since this is just a DTO for the line org, i'm not really sure the domain project is the correct place for this? Since it is only used one place, i would either use anonymous deserialization or this whole integration could be moved to the application project?

src/backend/api/Fusion.Resources.Api/Controllers/Person/PersonController.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using Fusion.AspNetCore.FluentAuthorization;
 using Fusion.Integration;
+using Fusion.Resources.Api.Controllers.Departments;
```
HansDahle, line 3: ?

src/backend/api/Fusion.Resources.Api/Controllers/Departments/DepartmentsController.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Threading.Tasks;
+
+namespace Fusion.Resources.Api.Controllers.Departments
+{
+    [ApiVersion("1.0-preview")]
+    [Authorize]
+    [ApiController]
+    public class DepartmentsController : ResourceControllerBase
+    {
+        [HttpGet("resources/departments/{departmentString}")]
```
HansDahle, line 16: I don't think we have any other uris formatted like this in the api? The other department related endpoints just use `/departments` on the root path.

src/backend/api/Fusion.Resources.Api/Controllers/Departments/DepartmentsController.cs
```diff
@@ -0,0 +1,40 @@
...
+                .ById(departmentString)
+                .ExpandResourceOwners();
+
+            var departments = await DispatchAsync(request);
+
+            return Ok(new ApiDepartment(departments.Single()));
+        }
+
+        [HttpGet("/resources/departments/search")]
+        public async Task<ActionResult<List<ApiDepartment>>> Search([FromQuery(Name = "q")] string query)
```
HansDahle, line 29: We have established the $search in all other apis. So should try to keep it consistent.

src/backend/api/Fusion.Resources.Api/Controllers/Departments/ApiDepartment.cs
```diff
@@ -0,0 +1,21 @@
...
+            Name = department.DepartmentId;
+            Sector = department.SectorId;
+            LineOrgResponsible = (department.LineOrgResponsible is not null) ? new ApiPerson(department.LineOrgResponsible) : null;
+            DefactoResponsible = (department.DefactoResponsible is not null) ? new ApiPerson(department.DefactoResponsible) : null;
+        }
+
+        public string Name { get; set; }
+        public string? Sector { get; set; }
+        public ApiPerson? LineOrgResponsible { get; set; }
+        public ApiPerson? DefactoResponsible { get; set; }
```
HansDahle, line 18: Is there a hard limit on just one delegated department responsible?

HansDahle (changes requested):

HansDahle:
> > Hmm.. the search endpoint uses search to the line org, then for each result dispatches a query for each department, which in turn makes a request to line org.. There will be some scaling issues here..
> 
> It only queries line org directly more than once when there are multiple pages. The other requests are to the people service, which I thought are cached?
> 
> > By caching the line org leaders, we will also insulate us a bit from transient errors. This could be cached ex. once a day etc.
> 
> True, but there are other trade-offs to consider when not operating on "live" data. Are we talking about caching in our database or setting up redis/something similar? The latter might be a bit much work for something that is likely to change in the near future.
> 

Using in memory caching should be fine for this. If we want to keep a record in our database i guess the term would be more in the lines of `replication and syncronization`. There are ~1700 line managers recorded in the line org db. If we resolve these, which should be fine in one page (ex. 2000), we can cache ex. just the ones in TPD+ -> using a model keeping only the props we need. This cache could be kept for a day ex or even just updated by the resolver, using a timer. The changes in the line org manager are slow. Ex. they do not reflect temporary changes etc.

src/backend/api/Fusion.Resources.Application/LineOrg/LineOrgResolver.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        public async Task<List<LineOrgDepartment>> GetResourceOwners(List<string> departmentIds, string? filter, CancellationToken cancellationToken)
+        {
+            var result = new List<LineOrgDepartment>();
+            var managedDepartments = new HashSet<string>(departmentIds);
+
+            var client = httpClientFactory.CreateClient("lineorg");
+
+            var uri = "/lineorg/persons?top=2000&$filter=isresourceowner eq true";
```
HansDahle, line 34: The query param is $top

HansDahle (approved):
You should make it a habit to review the files changed in github, so you can fix warnings in the touched files :)

Guess we shoulld just get this in so we get progress, however i think we can refactor this pretty soon, as the serach index has been updated with isResourceOwner etc.

PR #289, opened 2021-04-30
Resource Allocation Requests - ServiceBus topic notifications on request changes

xssic:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
Introducing fusion events, and dispatcing service bus topic notification for request changes (created and removed)

**Testing:**
- [ ] Can be tested
- [x] Local tests are passing

Tested by connecting to topic, and verifying flow to PIMS.

**Checklist:**
- [ ] Considered work items 
- [x] Considered security


src/backend/api/Fusion.Resources.Domain/Commands/Requests/InternalRequests/CreateInternalRequest.cs
```diff
@@ -132,6 +137,9 @@ private async Task<DbResourceAllocationRequest> CreateDbRequestAsync(CreateInter
                 };
 
                 dbContext.ResourceAllocationRequests.Add(item);
+                
+                await SendNotificationsAsync(new QueryResourceAllocationRequest(item));
```
HansDahle, line 141: Should try to align this with notifications. So the domain layer do not need to worry about the event integration etc.. Just publish an event that you handle in the api project, or even the application project i guess..

src/backend/api/Fusion.Resources.Domain/Events/ResourceAllocationRequestEventTypes.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+
+namespace Fusion.Events
+{
+    public static class ResourceAllocationRequestEventTypes
```
HansDahle, line 5: These models should not be defined in the domain model.. maybe they could be added to the Integration.Models?

HansDahle (changes requested):
Good start, but i think we should try and keep the domain project as clean as we can for integration.

src/backend/api/Fusion.Resources.Domain/Commands/Requests/InternalRequests/CreateInternalRequest.cs
xssic: dispatching from api project/controller

src/backend/api/Fusion.Resources.Domain/Events/ResourceAllocationRequestEventTypes.cs
xssic: Agree. Moved to Integration.Models

HansDahle (approved):

PR #299, opened 2021-05-06
Hide resource owner notes when request is completed

loldot:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
- Hide `GET` from options when request is completed to make it easier for frontend
- Delete comments when provisioning request.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security

### DIFF
--- a/src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
+++ b/src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
@@ -664,8 +664,7 @@ namespace Fusion.Resources.Api.Controllers
 
             if (authResult.Success)
             {
-                if (!request.IsCompleted) allowedMethods.Add("GET");
-                allowedMethods.Add("POST");
+                allowedMethods.Add("GET", "POST");
             }
 
             Response.Headers["Allow"] = string.Join(',', allowedMethods);
@@ -694,7 +693,8 @@ namespace Fusion.Resources.Api.Controllers
                 r.AlwaysAccessWhen().FullControl().FullControlInternal();
                 r.AnyOf(or =>
                 {
-                    or.BeResourceOwner(new DepartmentPath(requiredDepartment).Parent(), includeParents: true, includeDescendants: true);
+                    if (!request.IsCompleted)
+                        or.BeResourceOwner(new DepartmentPath(requiredDepartment).Parent(), includeParents: true, includeDescendants: true);
                 });
             });
             #endregion
@@ -703,8 +703,7 @@ namespace Fusion.Resources.Api.Controllers
 
             if (authResult.Success)
             {
-                if (!request.IsCompleted) allowedMethods.Add("GET");
-                allowedMethods.Add("PUT", "DELETE");
+                allowedMethods.Add("GET", "PUT", "DELETE");
             }
 
             Response.Headers["Allow"] = string.Join(',', allowedMethods);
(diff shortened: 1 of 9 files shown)


src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
```diff
@@ -654,11 +654,12 @@ public async Task<ActionResult> GetCommentOptions(Guid requestId)
                 r.AlwaysAccessWhen().FullControl().FullControlInternal();
                 r.AnyOf(or =>
                 {
-                    or.BeResourceOwner(new DepartmentPath(requiredDepartment).Parent(), includeParents: true, includeDescendants: true);
+                    if (!request.IsCompleted)
```
HansDahle, line 657: It kind of makes sense to hide it using 403, but i just think that down the line, this logic could be subject for confusion? Will we remember in 5 month why we are doing this check?

Maybe this should be handled separatly, instead of everything in one place, just to save lines?
IMO, i would either handle this before the normal auth check or after. 

```
if (request.IsComplete) {
    return new ObjectResult(new { error = new { 
              code = "CommentsUnavailable", 
              message = "Comments are not available on completed requests"
     } }) { StatusCode = 403 }
}
```
Maybe align with the default error model (rfc prob det.)... 
This would give info on why it's 403, not just that you don't have access. I often find myself wondring why the api respond how it does, when i expect otherwise, and often it is because i have just bundled some specific logic with other, just because it saved a few lines of code :)

src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/Allocation/ProvisionAllocationRequest.cs
```diff
@@ -46,6 +46,8 @@ protected override async Task Handle(ProvisionAllocationRequest request, Cancell
                         if (dbRequest.OrgPositionId is null)
                             throw new InvalidOperationException("Position id cannot be empty when provisioning request");
 
+                        await DeleteResourceOwnerNotes(request.RequestId, cancellationToken);
```
HansDahle, line 49: Would not a better place be to use the base Provision command? Here we are missing the resource owner requests and any other that might have comments.

Could add a `INotification` in the provision command, 

```
if (provisionSuccessfully)  {
  await mediator.Publish(new RequestProvisioned(requestId));
} else {
  await mediator.Publish(new RequestProvisioningFailed(requestId);
}

public DeleteNotesHandler : INotificationHandler<RequestProvisioned> 
{
  public async Task Handle(notification) {
     var comments = await GetCommentsAsync();
     db.RemoveRange(comments);
     db.SaveChanges();
  }
} 
```

The handler could just be in the logic library. But this event will be usefull eventually when we want to publish subscription events and send notifications...

HansDahle (changes requested):

src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
```diff
@@ -658,12 +658,13 @@ public async Task<ActionResult> GetCommentOptions(Guid requestId)
...
             });
             #endregion
-            
+
             var allowedMethods = new List<string> { "OPTIONS" };
 
             if (authResult.Success)
             {
-                allowedMethods.Add("GET", "POST");
+                if (!request.IsCompleted) allowedMethods.Add("POST");
```
HansDahle, line 666: Is the limitation that you cannot add new comments when it is completed?

HansDahle (approved):
Should be ok to push out, but there are a few suggestions that might be updated to improve consistency.

PR #313, opened 2021-05-16
Add auth for workflows

loldot:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security


src/backend/api/Fusion.Resources.Logic/Requests/CanApproveStepHandlerBase.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Threading.Tasks;
+
+namespace Fusion.Resources.Logic.Requests
+{
+    public record WorkflowAccessKey(string Subtype, string CurrentStep)
+    {
+        public static implicit operator WorkflowAccessKey((string, string) tuple)
+            => new WorkflowAccessKey(tuple.Item1, tuple.Item2);
+    }
+    public record WorkflowAccess()
```
HansDahle, line 18: Refactor to seperate file?

src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/Allocation/CanApproveStepHandler.cs
```diff
@@ -1,42 +1,79 @@
...
-using Fusion.Resources.Database;
 
 namespace Fusion.Resources.Logic.Commands
 {
     public partial class ResourceAllocationRequest
     {
-        public class CanApproveStepHandler : INotificationHandler<CanApproveStep>
+        public class CanApproveStepHandler : CanApproveStepHandlerBase, INotificationHandler<CanApproveStep>
         {
+            private static readonly Dictionary<WorkflowAccessKey, WorkflowAccess> AccessTable = new Dictionary<WorkflowAccessKey, WorkflowAccess>
```
HansDahle, line 19: Any chance we could move this into the workflow definition? I guess we can add that to a refactor task.. Maybe align with the fluent structure used there.

src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/Allocation/CanApproveStepHandler.cs
```diff
@@ -1,42 +1,79 @@
...
+                    .FirstAsync(r => r.Id == notification.RequestId, cancellationToken: cancellationToken);
 
-                //var initiatedBy = await dbContext.Persons.FirstAsync(p => p.Id == notification.InitiatedByDbPersonId);
-                //var request = await dbContext.ResourceAllocationRequests.FirstAsync(r => r.Id == notification.RequestId);
-
-                switch (notification.CurrentStepId)
-                {
-                    case "proposed":
+                var initiator = httpContextAccessor?.HttpContext?.User;
+                if (initiator is null) throw new UnauthorizedWorkflowException();
```
HansDahle, line 72: Should always include a message when throwing the same exception in different circumstances.

src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/Allocation/JointVentureAllocationRequestStarted.cs
```diff
@@ -56,10 +59,11 @@ private void ValidateWorkflow(DbResourceAllocationRequest request)
                             s.AddFailure("proposedPerson", "Must provide a person to be assigned the position"));
                 }
 
-                private ValueTask<bool> AssignRequestToResourceOwnerAsync()
+                private async ValueTask<bool> AssignRequestToResourceOwnerAsync(DbResourceAllocationRequest request)
                 {
-
-                    return new ValueTask<bool>(false);
+                    var department = await router.RouteAsync(request, CancellationToken.None);
```
HansDahle, line 64: Should not update this. Add this empty default failing method as a placeholder as the step was sort of unclarified from the spec.. But now it is more clear that the JV workflows should never have an approval step. 

But fixing the JV workflow should be done in a separate pr anyways..

src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/ResourceOwner/CanApproveResourceOwnerStepHandler.cs
```diff
@@ -0,0 +1,69 @@
...
+            public async Task Handle(CanApproveStep notification, CancellationToken cancellationToken)
+            {
+                if (notification.Type != DbInternalRequestType.ResourceOwnerChange) return;
+
+                var request = await dbContext.ResourceAllocationRequests
+                    .Include(p => p.Project)
+                    .FirstAsync(r => r.Id == notification.RequestId, cancellationToken: cancellationToken);
+
+                var initiator = httpContextAccessor?.HttpContext?.User;
+                if (initiator is null) throw new UnauthorizedWorkflowException();
```
HansDahle, line 61: Message

src/backend/api/Fusion.Resources.Logic/Requests/UnauthorizedWorkflowException.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Runtime.Serialization;
+
+namespace Fusion.Resources.Logic.Requests
+{
+    [Serializable]
+    public class UnauthorizedWorkflowException : Exception
```
HansDahle, line 7: Any point including all these constructors when they are never used? Is this som resharper magic template?

src/backend/api/Fusion.Resources.Logic/Workflows/AllocationJointVentureWorkflowV1.cs
```diff
@@ -69,6 +69,7 @@ public WorkflowStep SkipApproval()
             switch (current.Id)
             {
                 case CREATED:
```
HansDahle, line 71: Is this a fix for the workflow or is it regarding auth?

src/backend/api/Fusion.Resources.Logic/Requests/CanApproveStepHandlerBase.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                var result = await authorizationService.AuthorizeAsync(initiator, request, OrgProjectAccessRequirement.OrgWrite(request.Project.OrgProjectId));
+                isAllowed |= result.Succeeded;
+            }
+
+            isAllowed |= initiator.IsApplicationUser();
+            isAllowed |= initiator.IsInRole("Fusion.Resources.FullControl");
+            isAllowed |= initiator.IsInRole("Fusion.Resources.Internal.FullControl");
+
+            if (!isAllowed) throw new UnauthorizedWorkflowException();
```
HansDahle, line 120: Should include a message with which steps have been evaluated and the different outcomes? 
I think an informative message to the user could save us from much support requests. 

Should also add more telemetry here.. 
In the api we generate a payload based on the evaluated requirements. Maybe we could to that here as well. We should generate a similar payload and trace to AI, that way we can see whats going on _when_ we get questions.

src/backend/api/Fusion.Resources.Logic/Requests/CanApproveStepHandlerBase.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        public CanApproveStepHandlerBase(IAuthorizationService authorizationService)
+        {
+            this.authorizationService = authorizationService;
+        }
+        protected async Task CheckAccess(DbResourceAllocationRequest request, WorkflowAccess row, ClaimsPrincipal initiator)
+        {
+            bool isAllowed = false;
+
+            if (!string.IsNullOrEmpty(request.AssignedDepartment))
```
HansDahle, line 47: Could we use the fluent pattern here so we use the `if`'s to build the rule, then only evaluate once.. That way we only evaluate to the first success..

HansDahle (changes requested):

src/backend/api/Fusion.Resources.Logic/Requests/UnauthorizedWorkflowException.cs
loldot: It's from the default VS functionality when generating a class which ends with Exception. I guess the template is based on this advise: https://docs.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#include-three-constructors-in-custom-exception-classes

src/backend/api/Fusion.Resources.Logic/Workflows/AllocationJointVentureWorkflowV1.cs
loldot: It's related to the changes with request router.If there's never a need to approve joint venture requests, we should update the security matrix and revert this.

src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/Allocation/CanApproveStepHandler.cs
loldot: I'll refactor this in another branch.

HansDahle (approved):
Nice work.. Some warnings on nullability in ApiReportsTo that should be fixed. Should maybe also refactor to separate file?

I think we need to align the JV request with the enterprise request. But in another pr..

PR #326, opened 2021-06-01
Allow removing proposed person when request is not proposed yet

loldot:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
Allow unsetting proposed person when request is not sent for task owner approval.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security

### DIFF
--- a/src/backend/api/Fusion.Resources.Api/Controllers/Requests/Requests/PatchInternalRequestRequest.cs
+++ b/src/backend/api/Fusion.Resources.Api/Controllers/Requests/Requests/PatchInternalRequestRequest.cs
@@ -41,7 +41,7 @@ namespace Fusion.Resources.Api.Controllers
 
                         return false;
                     })
-                    .When(x => x.ProposedPersonAzureUniqueId.HasValue);
+                    .When(x => x.ProposedPersonAzureUniqueId.HasValue && x.ProposedPersonAzureUniqueId.Value.HasValue);
 
 
                 RuleFor(x => x.AssignedDepartment)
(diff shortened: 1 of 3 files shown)


src/backend/api/Fusion.Resources.Api/Controllers/Requests/Requests/PatchInternalRequestRequest.cs
```diff
@@ -41,7 +41,7 @@ public Validator(ResourcesDbContext db, IMediator mediator)
 
                         return false;
                     })
-                    .When(x => x.ProposedPersonAzureUniqueId.HasValue);
+                    .When(x => x.ProposedPersonAzureUniqueId.HasValue && x.ProposedPersonAzureUniqueId.Value.HasValue);
```
xssic, line 44: Reason for checking both fk & nav property ? (Just asking)

xssic (approved):

src/backend/api/Fusion.Resources.Api/Controllers/Requests/Requests/PatchInternalRequestRequest.cs
loldot: The first is to check whethert the value is set in the patch object at all, and the second is to check if the value is set to null when it is present in the patch object.

ie. { "someProp": "xx" } vs { "someProp": "xx", "proposedPersonAzureUniqueId": null }

In the first case we do not want to update the property, in the second we want to set the proposedPerson to null and cannot require that the person exists

PR #329, opened 2021-06-04
Department search: return items from line org when department not tracked in db

loldot:
- [ ] New feature
- [x] Bug fix
- [ ] High impact

**Description of work:**
Correctly check if search is limited to specific departments before filtering results from line org.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security

### DIFF
--- a/src/backend/tests/Fusion.Resources.Api.Tests/IntegrationTests/DepartmentsController.cs
+++ b/src/backend/tests/Fusion.Resources.Api.Tests/IntegrationTests/DepartmentsController.cs
@@ -140,6 +140,7 @@ namespace Fusion.Resources.Api.Tests.IntegrationTests
             var resp = await Client.TestClientGetAsync<TestDepartment>("/departments/TPD LIN ORG TST?api-version=1.0-preview");
 
             resp.Response.StatusCode.Should().Be(HttpStatusCode.OK);
+            resp.Value.Name.Should().Be("TPD LIN ORG TST");
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


src/backend/tests/Fusion.Resources.Api.Tests/IntegrationTests/DepartmentsController.cs
```diff
@@ -142,6 +142,36 @@ public async Task ShouldGetDepartmentNotInDbWhenInLineOrg()
...
+                        FullDepartment = "TPD LIN ORG TST"
+                    }
+                }
+            };
+
+            fixture.LineOrg.WithResponse("/lineorg/persons", lineorgData);
+
+            using var adminScope = fixture.AdminScope();
+
+            var resp = await Client.TestClientGetAsync<TestDepartment>($"/departments?$search={fakeResourceOwner.Name}&api-version=1.0-preview");
```
xssic, line 170: Hmmm. Guess the response may be right, but you are trying to return a single item here.
Search endpoint returns a list of departments. Or?

You could even check that the response value is the same as expected ;-)

xssic (changes requested):
Test might be right, but I am pretty sure it is returning the wrong model.

src/backend/tests/Fusion.Resources.Api.Tests/IntegrationTests/DepartmentsController.cs
loldot: Good catch

xssic (approved):

PR #336, opened 2021-06-10
Refactor department queries

loldot:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
- Refactor department queries for readability and use cases
- Fixes bug where adding a department to db would give conflict when it exists in line org.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security


src/backend/api/Fusion.Resources.Api/Controllers/Departments/DepartmentsController.cs
```diff
@@ -15,16 +16,17 @@ namespace Fusion.Resources.Api.Controllers.Departments
     [ApiController]
     public class DepartmentsController : ResourceControllerBase
     {
+        private readonly ResourcesDbContext db;
```
xssic, line 19: Don't think you should introduce db context here.
Should use query handler instead.

xssic (changes requested):
Not sure why you introduced DbContext in the controller, but I don't think it is the right way...

HansDahle (approved):

xssic (approved):

PR #361, opened 2021-07-08
Add conversations to requests

loldot:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
As described in https://github.com/equinor/fusion/pull/108/files?short_path=3651bf1#diff-3651bf19d4481f8b35676b13cc20ce0fceab59f6da5d465bb741c82461a7e093

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security

### DIFF
--- a/src/backend/api/Fusion.Resources.Database/Entities/DbResourceAllocationRequest.cs
+++ b/src/backend/api/Fusion.Resources.Database/Entities/DbResourceAllocationRequest.cs
@@ -70,6 +70,7 @@ namespace Fusion.Resources.Database.Entities
 
         public DbOpProvisionStatus ProvisioningStatus { get; set; } = new DbOpProvisionStatus();
         public List<DbRequestTask>? Tasks { get;  set; }
+        public List<DbConversationMessage> Conversation { get; set; } = new();
 
         internal static void OnModelCreating(ModelBuilder modelBuilder)
         {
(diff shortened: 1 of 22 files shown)


src/backend/api/Fusion.Resources.Database/Entities/DbResourceAllocationRequest.cs
```diff
@@ -70,6 +70,7 @@ public class DbResourceAllocationRequest
 
         public DbOpProvisionStatus ProvisioningStatus { get; set; } = new DbOpProvisionStatus();
         public List<DbRequestTask>? Tasks { get;  set; }
+        public List<DbConversationMessage> Conversation { get; set; } = new();
```
HansDahle, line 73: Is there a reason we new up this property, but not the others?

HansDahle (changes requested):
Mostly good stuff, but we cannot go live with this with no form of authorization. 
I think we must use the different entrypoints for fetching a request to indicate authorization/filtering. 

Doing this in a generic endpoint, we would have to identify why the user can access the request. I guess we could do that by running two authorization sequences, but would be more complex... 

Could this be added as an $expand to fetching a single request?

HansDahle (approved):

PR #364, opened 2021-07-09
Reset allocation state/workflow for request

loldot:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
Add endpoint to reset state and workflow when sending request back to task owner.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security

### DIFF
--- a/src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
+++ b/src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
@@ -700,6 +700,25 @@ namespace Fusion.Resources.Api.Controllers
             if (requestItem == null)
                 return ApiErrors.NotFound("Could not locate request", $"{requestId}");
 
+            #region Authorization
+
+            var authResult = await Request.RequireAuthorizationAsync(r =>
+            {
+                r.AlwaysAccessWhen().FullControl().FullControlInternal();
+                r.AnyOf(or =>
+                {
+                    if (!string.IsNullOrEmpty(requestItem.AssignedDepartment))
+                        or.BeResourceOwner(new DepartmentPath(requestItem.AssignedDepartment).Parent(), includeParents: false, includeDescendants: true);
+                    else
+                        or.BeResourceOwner();
+                });
+            });
+
+            if (authResult.Unauthorized)
+                return authResult.CreateForbiddenResponse();
+
+            #endregion
+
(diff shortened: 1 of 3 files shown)


src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
```diff
@@ -691,6 +691,39 @@ public async Task<ActionResult<ApiResourceAllocationRequest>> ApproveProjectAllo
...
+            });
+
+            if (authResult.Unauthorized)
+                return authResult.CreateForbiddenResponse();
+
+            #endregion
+
+            var deleted = await DispatchAsync(new ResetWorkflow(requestId));
+            if (deleted) return NoContent();
+            else return NotFound();
```
HansDahle, line 724: Kind of ambigious. Could interpret this as the request does not exist, not the workflow. 
It could go either way if 404 is relevant, just delete it if it exists. In essense what is done here is setting a property to null on the request. 
In the least if you want to return a not found, a message should be included.

src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
```diff
@@ -691,6 +691,39 @@ public async Task<ActionResult<ApiResourceAllocationRequest>> ApproveProjectAllo
...
+            if (requestItem == null)
+                return ApiErrors.NotFound("Could not locate request", $"{requestId}");
+
+            #region Authorization
+
+            var authResult = await Request.RequireAuthorizationAsync(r =>
+            {
+                r.AlwaysAccessWhen().FullControl().FullControlInternal();
+                r.AnyOf(or =>
+                {
```
HansDahle, line 709: Should anyone on the taskowner side be able to reset the workflow?

HansDahle (approved):
Looks ok, few non critical comments.. Though in general, when returning a non successfull response, always include a message.

src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
loldot: I thought about this, but couldn't think of any use cases where that would be required and decided to put it off for now, since we can always add that in easily later, while it's a bit hard to think about the implications of letting the task owner side delete the workflow.

PR #388, opened 2021-08-31
Add endpoint to remove delegated resource owner

loldot:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
Add endpoint to delete delegated resource owners

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security

### DIFF
--- a/src/backend/api/Fusion.Resources.Api/Controllers/Departments/DepartmentsController.cs
+++ b/src/backend/api/Fusion.Resources.Api/Controllers/Departments/DepartmentsController.cs
@@ -96,7 +96,7 @@ namespace Fusion.Resources.Api.Controllers
                 new DeleteDelegatedResourceOwner(departmentString, azureUniqueId)
             );
 
-            if (deleted) return Ok();
+            if (deleted) return NoContent();
             else return NotFound();
         }
 
--- a/src/backend/api/Fusion.Resources.Api/Controllers/Departments/DepartmentsController.cs
+++ b/src/backend/api/Fusion.Resources.Api/Controllers/Departments/DepartmentsController.cs
@@ -79,6 +79,27 @@ namespace Fusion.Resources.Api.Controllers
             return CreatedAtAction(nameof(GetDepartments), new { departmentString }, null);
         }
 
+        [HttpDelete("/departments/{departmentString}/delegated-resource-owner/{azureUniqueId}")]
+        public async Task<IActionResult> DeleteDelegatedResourceOwner(string departmentString, Guid azureUniqueId)
+        {
+            #region Authorization
+            var authResult = await Request.RequireAuthorizationAsync(r =>
+            {
+                r.AlwaysAccessWhen().FullControl().FullControlInternal();
+            });
+
+            if (authResult.Unauthorized)
+                return authResult.CreateForbiddenResponse();
+            #endregion
+
+            var deleted = await DispatchAsync(
+                new DeleteDelegatedResourceOwner(departmentString, azureUniqueId)
+            );
+
+            if (deleted) return Ok();
+            else return NotFound();
+        }
+
         [HttpGet("/projects/{projectId}/positions/{positionId}/instances/{instanceId}/relevant-departments")]
         public async Task<ActionResult<ApiRelevantDepartments>> GetPositionDepartments(
             Guid projectId, Guid positionId, Guid instanceId)
--- a/src/backend/api/Fusion.Resources.Domain/Commands/Departments/DeleteDelegatedResourceOwner.cs
+++ b/src/backend/api/Fusion.Resources.Domain/Commands/Departments/DeleteDelegatedResourceOwner.cs
@@ -12,12 +12,12 @@ namespace Fusion.Resources.Domain.Commands.Departments
     public  class DeleteDelegatedResourceOwner : IRequest<bool>
     {
         private readonly string departmentId;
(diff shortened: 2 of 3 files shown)


src/backend/api/Fusion.Resources.Domain/Commands/Departments/DeleteDelegatedResourceOwner.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Fusion.Resources.Domain.Commands.Departments
+{
+    public  class DeleteDelegatedResourceOwner : IRequest<bool>
+    {
+        private readonly string departmentId;
+        private readonly Guid delegatedOnwerAzureUniqueId;
```
xssic, line 15: should probably be owner (and not onwer)

src/backend/api/Fusion.Resources.Domain/Commands/Departments/DeleteDelegatedResourceOwner.cs
```diff
@@ -0,0 +1,43 @@
...
+            private readonly ResourcesDbContext db;
+
+            public Handler(ResourcesDbContext db)
+            {
+                this.db = db;
+            }
+
+            public async Task<bool> Handle(DeleteDelegatedResourceOwner request, CancellationToken cancellationToken)
+            {
+                var query = db.DepartmentResponsibles
```
xssic, line 34: I prefer to query for items, and remove(range) if found, then only call savechanges if actually modifying the database

src/backend/api/Fusion.Resources.Api/Controllers/Departments/DepartmentsController.cs
```diff
@@ -79,6 +79,27 @@ public async Task<ActionResult> AddDelegatedResourceOwner(string departmentStrin
...
+
+            if (authResult.Unauthorized)
+                return authResult.CreateForbiddenResponse();
+            #endregion
+
+            var deleted = await DispatchAsync(
+                new DeleteDelegatedResourceOwner(departmentString, azureUniqueId)
+            );
+
+            if (deleted) return Ok();
```
xssic, line 99: I would return NoContent instead of OK

src/backend/api/Fusion.Resources.Api/Controllers/Departments/DepartmentsController.cs
```diff
@@ -79,6 +79,27 @@ public async Task<ActionResult> AddDelegatedResourceOwner(string departmentStrin
...
+            else return NotFound();
```
xssic, line 100: Any possibility to check for existing data and return NotFound before issuing DeleteDelegatedResourceOwner ?

I would have checked for data before auth check returning NotFound if nothing to remove.

src/backend/api/Fusion.Resources.Domain/Commands/Departments/DeleteDelegatedResourceOwner.cs
xssic: if(query.Any){removerange, savechanges}

src/backend/tests/Fusion.Resources.Api.Tests/IntegrationTests/DepartmentsController.cs
```diff
@@ -110,6 +110,44 @@ public async Task AddDepartmentResponsible_ShouldBeAllowed_WhenAdmin()
...
+            {
+                DateFrom = "2021-02-02",
+                DateTo = "2022-02-05",
+                ResponsibleAzureUniqueId = fakeResourceOwner.AzureUniqueId
+            });
+
+            resp = await Client.TestClientDeleteAsync<dynamic>(
+                $"/departments/{testDepartment}/delegated-resource-owner/{fakeResourceOwner.AzureUniqueId}?api-version=1.0-preview"
+            );
+            resp.Response.StatusCode.Should().Be(HttpStatusCode.OK);
```
xssic, line 133: NoContent?

src/backend/api/Fusion.Resources.Domain/Commands/Departments/DeleteDelegatedResourceOwner.cs
loldot: Wouldn't that require two roundtrips to the database? (I think the call to .Any() enumerates the enumerable). Whereas the method used here will only do one roundtrip to the database. Sending delete x where y when running savechangesasync().

src/backend/api/Fusion.Resources.Api/Controllers/Departments/DepartmentsController.cs
loldot: I don't think disclosing whether an item exists or not to unauthorized users is necessary, in most cases it does not matter, but in some cases even knowing something is there could be sensitive info.

src/backend/api/Fusion.Resources.Domain/Commands/Departments/DeleteDelegatedResourceOwner.cs
xssic: probably ok, just asking :-)

xssic (approved):

PR #409, opened 2021-10-19
Api - Analytics - Add person reference to absence analytics endpoint

xssic:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
Add person model to absence analytics endpoint.
Since we query for all absence, we require person info to map against.

[AB#22603](https://statoil-proview.visualstudio.com/9949ad5e-7d15-4531-901e-296574079ee1/_workitems/edit/22603)

**Testing:**
- [ ] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [ ] Considered security

### DIFF
--- a/src/backend/api/Fusion.Resources.Api/Controllers/Requests/AnalyticsController.cs
+++ b/src/backend/api/Fusion.Resources.Api/Controllers/Requests/AnalyticsController.cs
@@ -99,7 +99,7 @@ namespace Fusion.Resources.Api.Controllers
             public Guid Id { get; set; }
             public bool IsPrivate { get; set; }
             public string? Comment { get; set; }
-            public ApiPerson Person { get; set; }
+            public ApiPerson? Person { get; set; }
             public ApiTaskDetails? TaskDetails { get; set; }
             public DateTimeOffset AppliesFrom { get; set; }
             public DateTimeOffset? AppliesTo { get; set; }


src/backend/api/Fusion.Resources.Api/Controllers/Requests/AnalyticsController.cs
```diff
@@ -85,6 +85,8 @@ private ApiPersonAbsenceForAnalytics(QueryPersonAbsenceBasic absence)
                 Comment = absence.Comment;
                 TaskDetails = (absence.TaskDetails != null) ? new ApiTaskDetails(absence.TaskDetails) : null;
 
+                Person = new ApiPerson(absence.Person!);
```
HansDahle, line 88: Maybe rather do a nullcheck and throw your own error, instead of this potentially resulting in a nullreferror.. 
Or just do a check, if not null, set the property.. 

Kind of depends what we want as an outcome.. Do we want 1 row failing, or the whole dataset failing, if one row is invalid... ?

xssic: All rows will have a person. But null check will be added.

HansDahle (approved):

PR #432, opened 2021-11-03
Fixed filtering av employment statuses where applies to date could be…

jkjaervik:
… null


jkjaervik:
- [ ] New feature
- [x] Bug fix
- [ ] High impact

*Description of work:*
Abscences/Employment status where AppliesToDate was null would not get included when expanding with timeline. Added a check on assigned person on logic for change requests on a position.

- [x] Can be tested
- [x] Automatic test created / updated
- [x] Local tests are passing

HansDahle (approved):
lgtm

PR #446, opened 2021-12-13
fix: table scroll fix and overdue icon

eskil0312:
- Make sure only table scrolls, not entire page
- Added warning  icon to actual mpp page of overdue or soon overdue 
NOTE: Use fusion-components version of warning icon. Tried with fusion react icon, but got error.
Use fusion comp version for now due to the high propriety of this pr

### DIFF
--- a/src/frontend/package-lock.json
+++ b/src/frontend/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "resources",
-  "version": "2.1.5",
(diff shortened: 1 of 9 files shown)


terjebra (approved):
Looks good

tobiasny (commented):
Changes look fine to me :) @terjebra will you have a final review and approve these changes?

PR #453, opened 2022-01-14
API - ExternalPersonnel handle/expose properties indicating users deleted in azure

xssic:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
* Added DELETED property to DBExternalPersonnel
* Added IsDeleted & DELETED to query models (QueryContractPersonnel/QueryExternalPersonnelPerson)
* Added IsDeleted & DELETED to api models/odata filters (ApiContractPersonnel/ApiExternalPersonnelPerson)
* Added DELETED to be updated during profile refresh (IsDeleted already being updated)

[AB#25133](https://statoil-proview.visualstudio.com/9949ad5e-7d15-4531-901e-296574079ee1/_workitems/edit/25133)

**Testing:**
- [ ] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [ ] Considered security


loldot (approved):

HansDahle:
What should we do regarding already deleted accounts.. 
Just leave the deleted date as null? or set today? 

I think this date should exist in the people api, if you resolve the account given the azure id. Maybe in the profile refresh, we can use this property if the account is deleted, but the deleted date is null?

xssic:
I will update deleted date based on PPL-expired date for already existing accounts.
Current implementation resolves by mail, but we could probably extend the profile sync to resolve by azure unique id when available. 
To be able to use expired-flags, the Fusion.Integration.Profile.Abstractions must be updated.

PR #472, opened 2022-02-18
 Flag deleted instances

loldot:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
- [AB#26367](https://statoil-proview.visualstudio.com/9949ad5e-7d15-4531-901e-296574079ee1/_workitems/edit/26367)

Sometimes a request is created for a position instance and then later the position instance is deleted in org.
This change should flag when that has happened.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security

### DIFF
--- a/src/backend/api/Fusion.Resources.Domain/Models/QueryResourceAllocationRequest.cs
+++ b/src/backend/api/Fusion.Resources.Domain/Models/QueryResourceAllocationRequest.cs
@@ -135,6 +135,8 @@ namespace Fusion.Resources.Domain
         public List<QueryRequestAction>? Actions { get; set; }
         public List<QueryConversationMessage>? Conversation { get; set; }
 
+        public bool IsOrgInstanceDeleted => OrgPositionInstanceId.HasValue && OrgPositionInstance == null;
+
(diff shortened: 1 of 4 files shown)


src/backend/api/Fusion.Resources.Domain/Models/QueryResourceAllocationRequest.cs
```diff
@@ -135,6 +135,8 @@ public QueryResourceAllocationRequest(DbResourceAllocationRequest entity, QueryW
         public List<QueryRequestAction>? Actions { get; set; }
         public List<QueryConversationMessage>? Conversation { get; set; }
 
+        public bool IsOrgInstanceDeleted => OrgPositionInstanceId.HasValue && OrgPositionInstance == null;
```
HansDahle, line 138: Are we sure OrgPositionInstance allways will be present? 
I was thinking more of hooking in to the event stream from org and update the db with a property when an instance is deleted. 
That way we can crawl for corrupt data and use it in queries etc..

PR #476, opened 2022-03-01
Key rotation pipeline update

HansDahle:
Updated the pipeline to use latest version of powershell, which allows setting a display name for the secrets generated. This leads to improved visibility when inspecting expiring secrets etc.

Issues encountered due to bugs in the PS module at version 7.1.0, which is default on hosted agents. Resolved this by forcing 7.2.0 version which is normally pre-installed on the agents.


xssic (approved):
nice 👍 

Do you know if we could/should do similar exercise for other deploy pipelines too ?

PR #493, opened 2022-05-19
Pro-Org - Positions having allocation request missing in org-chart histogram

xssic:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
When creating or deleting request, logic should update hasRequest property on pro-org position instance.
true if request is created
false if request is deleted

Property does not exist if no request exists. Property may exist on older instances created based on PIMS as source.
(Previously requests where handled in PIMS)

AB#28770

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

Should be able to verify using histograms in Pro-Org app.
Creating request, should tag instance. Should reflect in historgram's.

**Checklist:**
- [x] Considered work items 
- [x] Considered security

### DIFF
--- a/src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/UpdateOrgPositionInstanceHaveRequest.cs
+++ b/src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/UpdateOrgPositionInstanceHaveRequest.cs
@@ -5,50 +5,48 @@ using System;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using Fusion.Resources.Domain.Queries;
 
 namespace Fusion.Resources.Logic.Commands
 {
     public class UpdateOrgPositionInstanceHaveRequest : IRequest
     {
-        public UpdateOrgPositionInstanceHaveRequest(Guid requestId, bool haveRequest)
+        public UpdateOrgPositionInstanceHaveRequest(Guid orgProjectId, Guid orgPositionId, Guid orgPositionInstanceId, bool haveRequest)
         {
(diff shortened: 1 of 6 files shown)


src/backend/api/Fusion.Resources.Logic/Requests/Commands/ResourceAllocationRequest/UpdateOrgPositionInstanceHaveRequest.cs
```diff
@@ -0,0 +1,73 @@
...
+using Newtonsoft.Json.Linq;
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Fusion.Resources.Domain.Queries;
+
+namespace Fusion.Resources.Logic.Commands
+{
+    public class UpdateOrgPositionInstanceHaveRequest : IRequest
```
loldot, line 12: We already have notifications for `InternalRequestCreated` and `InternalRequestDeleted`. Could add a handler for both these notifications and set hasRequest = true false based on type of notification instead of adding a new request issued by controllers. I.e. 

```cs
UpdateOrgPositionInstanceHaveRequestHandler : INotificationHandler<InternalRequestCreated>, INotificationHandler<InternalRequestDeleted>
{
     Handle(InternalRequestCreated...) => Update(..., hasRequest: true);
     Handle(InternalRequestDeleted...) => Update(..., hasRequest: false);
}
```

xssic: Could utilize this yes. Agree :-)
Revoved from controller

loldot (approved):

PR #509, opened 2022-07-26
Enable support for usage of org unit scoped roles in authorization requirements

loldot:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
AB#30484

Resource owners are currently limited to accessing competency centers related to their location in line org. Certain scenarios exist where it would be useful to grant resource owner access to competency centers even though the regular line org hierarchy should not allow it. I.e. EPN SUB and EPI SUB shares resources and would like to be able to get an overview of all competency centers. The current `FullControl` role is too coarse grained to be used extensively, therefore this PR adds a new OrgUnit scoped role, `ResourceOwner`.

In contrast to the permissions granted from being a resource owner in line org, the new role allows being explicit about how the access is granted with regards to subdepartments. 

i.e. when a person is resource owner in PDP PRD FE, he or she will have write access to all departments and their subdepartments in PDP PRD FE, and read access to all departments and subdepartments in PDP PRD.

Granting the new scoped role for resource owner to PDP PRD FE will only grant read/write access to PDP PRD FE, not any parents or subdepartments. To grant access to subdepartments, the scope PDP PRD FE * can be used instead.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items  ([AB#30484](https://statoil-proview.visualstudio.com/Fusion%20Apps%20-%20Product%20Owners/_backlogs/backlog/Equinor%20Resource%20Management/Stories/?workitem=30484))
- [x] Considered security


src/backend/Fusion.Resources.Authorization/Fusion.Resources.Authorization.csproj
```diff
@@ -13,9 +13,9 @@
     </Content>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Fusion.ApiClients.Org" Version="5.0.6" />
+    <!--<PackageReference Include="Fusion.ApiClients.Org" Version="5.0.6" />-->
```
xssic, line 16: If not needed, remove

src/backend/api/Fusion.Resources.Api/Fusion.Resources.Api.csproj
```diff
@@ -19,14 +19,14 @@
 
 	<ItemGroup>
 		<PackageReference Include="Bogus" Version="34.0.2" />
-		<PackageReference Include="Fusion.Integration" Version="6.0.0" />
+		<PackageReference Include="Fusion.Integration" Version="6.0.3" />
 		<PackageReference Include="Fusion.Events.Server" Version="3.0.0" />
 		<!-- No longer need this reference, but keeping for future ref. -->
 		<!--<PackageReference Include="Fusion.Integration.Profile.Abstractions" Version="5.1.0" />-->
```
xssic, line 25: Not yours to change, but not sure why we need this for future ref... (ref. the one I asked to remove)

src/backend/api/Fusion.Resources.Domain/Commands/Departments/AddDelegatedResourceOwner.cs
```diff
@@ -24,25 +20,24 @@ public AddDelegatedResourceOwner(string departmentId, Guid responsibleAzureUniqu
...
                 {
-                    DateCreated = DateTime.UtcNow,
-                    DateFrom = request.DateFrom,
-                    DateTo = request.DateTo,
-                    DepartmentId = request.DepartmentId,
-                    ResponsibleAzureObjectId = request.ResponsibleAzureUniqueId,
-                };
-
-                db.DepartmentResponsibles.Add(delegatedResourceOwner);
-                await db.SaveChangesAsync(cancellationToken);
```
xssic, line 45: Just asking..
This change is intended ? (to move from resource-db-table to roles service?
If this table is not to be used anymore, we should remove it.

src/backend/tests/Fusion.Testing.Mocks.ProfileService/Api/PersonsController.cs
```diff
@@ -19,6 +19,7 @@ namespace Fusion.Testing.Mocks.ProfileService.Api
     public class PersonsController : ControllerBase
     {
         [MapToApiVersion("3.0")]
```
xssic, line 21: in this test, do we need the old api-version ?

xssic (commented):
Guess you should fix all failing tests first.
Maybe also remove table DepartmentResponsibles and refs, if not to be used anymore ?

src/backend/api/Fusion.Resources.Domain/Commands/Departments/AddDelegatedResourceOwner.cs
loldot: yes, the change is intended. We could remove the table, but it's still in use when checking access. Removing the table would require a very complex migration from the table to the roles service. I would think it's better to wait untill all delegations have expired and delete at a later stage.

src/backend/tests/Fusion.Testing.Mocks.ProfileService/Api/PersonsController.cs
loldot: yes, but new version is not required

xssic (approved):

xssic (changes requested):
Do believe we need to revise role-integration

xssic (approved):

PR #560, opened 2023-02-08
Direct request only option on base position

HansDahle:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
Some business areas anchor their base positions high up in the organisation as there are multiple paths for resources and no clear ownership.

To support these scenarios a new setting was introduced on the base position, which indicates that normal requests should not be created as these would be assigned to a leader that would not respond to it.

This option should be respected by the resources api when creating requests, not allowing these to be created. 

The api has been updated to resolve the base position for the position the request is created on and checks if the setting is present. If the flag is set to `true` the request is rejected if the sub type is not set to `direct`.

Caveats:
- Used the command validation pipeline to do the validation to hook in to the resolving of subtype. However this project does not have reference to the logic project where the constants live. Instead of refactoring where constants exists a string is used.
- Response out of api after command validation fails results in `BadRequest` however the description is not optimal, however by reading the body one can resolve what is the problem.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

Created tests to verify bad request is returned when creating a normal allocation request when the base position has `requireDirectRequest` flag.
~~Created test to verify auto resolving to direct request when no sub type is specified in body~~ (Removed this functionality)
Created test to verify direct request can be created when base position has flag.

Should test how the UI will behave when creating a new request without specifying a person (direct request). Unsure if the frontend will handle the bad request response.

**Checklist:**
- [x] Considered work items 
- [x] Considered security

[AB#36031](https://statoil-proview.visualstudio.com/9949ad5e-7d15-4531-901e-296574079ee1/_workitems/edit/36031)

Should update docs to specify functionality.

### DIFF
--- a/src/backend/tests/Fusion.Resources.Api.Tests/IntegrationTests/InternalRequests/DirectRequestTests.cs
+++ b/src/backend/tests/Fusion.Resources.Api.Tests/IntegrationTests/InternalRequests/DirectRequestTests.cs
@@ -154,6 +154,50 @@ namespace Fusion.Resources.Api.Tests.IntegrationTests
         }
 
         [Fact]
+        public async Task DirectRequest_Create_ShouldBeSuccessfull_WhenDirectRequestRequiredAndSubTypeDirect()
+        {
+            using var adminScope = fixture.AdminScope();
+            var directRequestBasePosition = testProject.AddBasePosition($"Direct request test {Guid.NewGuid()}", s =>
+            {
+                s.Settings["requireDirectRequest"] = true;
+            });
+
+            var position = testProject.AddPosition(p => p.BasePosition = directRequestBasePosition);
+            var proposedPerson = fixture.AddProfile(FusionAccountType.Employee);
+
+            var response = await Client.TestClientPostAsync<TestApiInternalRequestModel>($"/projects/{projectId}/requests", new
+            {
+                type = "normal",
+                subType = "direct",
+                orgPositionId = position.Id,
+                orgPositionInstanceId = position.Instances.Last().Id,
+            });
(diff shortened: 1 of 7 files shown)


oivheg (commented):
One of the tests fails, other than that the code looks fine.  se specific comment.

The other test that fail, do usualy fix them selfes after a couple of runs.  are something in the test run setup that happens sometimes.

src/backend/tests/Fusion.Resources.Api.Tests/IntegrationTests/InternalRequests/DirectRequestTests.cs
```diff
@@ -223,6 +267,27 @@ public async Task DirectRequest_Create_ShouldBeAbleToProposePersonDirectly()
...
+                s.Settings["requireDirectRequest"] = true;
+            });
+
+            var position = testProject.AddPosition(p => p.BasePosition = directRequestBasePosition);
+            var proposedPerson = fixture.AddProfile(FusionAccountType.Employee);
+
+            var response = await Client.TestClientPostAsync<TestApiInternalRequestModel>($"/projects/{projectId}/requests", new
+            {
+                type = "normal",
+                orgPositionId = position.Id,
```
xssic, line 285: When you don't provide a subType, the proper subtype will be resolved(in ResolveSubType).
To make this test green, you have to provide an invalid subType

xssic (commented):
Code seems fine, but you have to adjust the test, providing an invalid subType.
Not providing subtype will provide the proper subtype.

oivheg (approved):

PR #572, opened 2023-05-30
feat(api): Ensure data is committed to database before public events are dispatched

HansDahle:
- [x] New feature
- [x] Bug fix
- [ ] High impact

**Description of work:**
Noticed events are dispatched on event bus before data is commited to database. This means external parties can react and query data before this is available.

Fixed by utilizing the transaction feature available in fusion event library. This will store events internally on a queue in memory and only dispatch on the event bus when the transaction is commited.
Created new transaction object that wraps both db and event transaction.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

No new tests created. Current endpoints verified by existing tests. 

**Checklist:**
- [x] Considered security


oivheg (approved):

HansDahle:
Verified delete request works, events are dispatched through event bus. 
Verified create new request works, events are dispatched through event bus at the same time as response is returned, after a total of 4s request time. The long delay should be the interaction with the org api, indicating that the transaction scope works as expected.

HansDahle:
Since the transaction feature in event lib uses AsyncLocal the scope has to be created in the same async method as the rest of the code executes. If this is done in a  'branch' the AsyncLocal scope will be cleared when continuing to the next statement in the method.

PR #582, opened 2023-07-06
Updated the PeopleSearchIndex Query to support same manager on multiple Departments at same Level

oivheg:
- [ ] New feature
- [X] Bug fix
- [ ] High impact

**Description of work:**

When a User is manager in more than one Departmetn at the same Level, then the search results returnes the personnel from all departments the manager is managing. 
To help mitigate this the code is chagned to also include "isResourceOwner eq true"

This should help remove the problem with the personnell showing in wrong departments. 

Ther are howere cases we need to think  about if is applicable. 

- Should DelegatedResposibility users also be included?
- Should whe check for the role Fusion.ResourseOwner and include them? do they appear as managers in eq One Equinor?
- this appears to be Identical code that are in **LineOrg Service** file  **ListOrgUnitEmployees.cs**

[AB#41694](https://statoil-proview.visualstudio.com/9949ad5e-7d15-4531-901e-296574079ee1/_workitems/edit/41694)

**Testing:**
- [ ] Can be tested
- [ ] Local tests are passing

AB#
**Checklist:**
- [ ] Considered work items 
- [ ] Considered security


HansDahle (approved):
This solves the problem. GJ.

PR #636, opened 2024-05-30
feat: Remove legacy lineorg endpoint usage

HansDahle:
- [x] New feature
- [ ] Bug fix
- [x] High impact

**Description of work:**
Removed all usage of the `/lineorg/persons` and `/department` endpoints for internal logic related to org units. 
Endpoints should mostly return the same results, however the new data set for org units are more consistent so could result in some variations.

- Removed usage of `/lineorg/persons` to resolve resource owner
- Removed usage of `/departments` in lineorg service to list org units and related departments (siblings/children etc). 
- Refactored query model to be more semantic correct, which caused changes in a few additional files.
- Added some properties to response model to prepare for usage of sap id on frontend. 

- Updated mock setup and infra as we are using new endpoints and more structured org unit data. Should refactor this into `org-unit-builder` functionality in future pr. Omitting too many changes to keep scope down in pr.
 
**Testing:**
- [x] Can be tested
- [x] Local tests are passing

Fixed all failing tests by updating setup logic. Discovered some implementation errors by failing tests as well. 

**Checklist:**
- [x] Considered work items 
- [x] Considered security

[AB#53664](https://statoil-proview.visualstudio.com/787035c2-8cf2-4d73-a83e-bb0e6d937eec/_workitems/edit/53664)


src/backend/api/Fusion.Resources.Domain/Commands/Departments/GetRelatedDepartments.cs
```diff
@@ -61,19 +52,38 @@ public Handler(ILogger<Handler> logger, IMediator mediator, ILineOrgResolver lin
...
-                if(siblings is not null)
+                var orgUnit = await lineOrg.ResolveOrgUnitAsync(fullDepartmentPath, a => a.ExpandChildren());
+
+                if (orgUnit is null)
+                    return null;
+
+                // Add children
+
+                // Children should be returned
+                if (relevantDepartments.Children is null)
```
aleklundeq, line 65: Should the relevantDepartment always have children? Even if the relevantdepartment we check is at the "bottom"?

src/backend/api/Fusion.Resources.Domain/Models/QueryDepartment.cs
```diff
@@ -1,25 +1,58 @@
...
-            LineOrgResponsible = manager;
-        }
+            FullDepartment = lineOrgUnit.FullDepartment;
+            Identifier = lineOrgUnit.SapId;
+            Name = lineOrgUnit.Name;
+            ShortName = lineOrgUnit.ShortName;
+
+            if (lineOrgUnit.Management is not null)
+            {
+                var profile = lineOrgUnit.Management.Persons.FirstOrDefault();
```
aleklundeq, line 20: Would it be possible to have more than one persons in management-list? And why is the first one set as "LineOrgResponsible"?

aleklundeq (approved):

src/backend/api/Fusion.Resources.Domain/Commands/Departments/GetRelatedDepartments.cs
HansDahle: If this is set to null it indicates that the server did not include it in it's response, which would either be an error on their site or we have some issue internally where a cached copy is used which has not been fetched correctly..

src/backend/api/Fusion.Resources.Domain/Models/QueryDepartment.cs
HansDahle: Yes. This is a short term quick fix, and should work for most departments. However we need to address handling multiple managers, but this must also be supported in the UI. As there is only one property on the api model.

PR #639, opened 2024-06-04
feat: Update assigned department on request when org unit is renamed

HansDahle:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
When org units are renamed or moved we still persist the old full department string. This can leave requests connected to org units that no longer exists -> orphan requests.

We can subscribe for events comming from line org to act when these changes are done. This should be done as a persistant event handler so we are guaranteed delivery in case of downtime etc.

- Added persistant event handler to perform sync on assigned department for requests

The transient handler are used to invalidate cache, as this is executed on all instances of the api.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

Can be verified by assigning a ticket to some department (use one not commonly used, so we dont cause much problems).
PATCH the org unit in lineorg, chaning the full department (add a character etc).
Verify requests are assigned to that department
PATCH the org unit back to the original value (clean up)
Can verify requests are assigned back

**Checklist:**
- [ ] ~~Considered automated tests~~
- [x] Considered work items 
- [x] Considered security

[AB#53786](https://statoil-proview.visualstudio.com/787035c2-8cf2-4d73-a83e-bb0e6d937eec/_workitems/edit/53786)
Limited changes to only requests. There are other entities which stores full department reference. These should be changed in own prs.

### DIFF
--- a/src/backend/api/Fusion.Resources.Api/FusionEvents/LineOrgSyncronizationHandler.cs
+++ b/src/backend/api/Fusion.Resources.Api/FusionEvents/LineOrgSyncronizationHandler.cs
@@ -1,0 +1,81 @@
+﻿using Fusion.Events;
+using System.Threading;
+using System.Threading.Tasks;
+using Fusion.Resources.Domain;
+using Newtonsoft.Json;
+using MediatR;
+using System;
+using System.Linq;
+
+namespace Fusion.Resources.Api
+{
+    /// <summary>
+    /// Handler to execute syncronisation events from the line org service. 
+    /// We only want updates to be executed by a single instance in case of multiple instances running the api. 
+    /// 
+    /// The handler should accept events, collect relevant data, determine if event is relevant and dispatch events the domain should handle to perform updates.
+    /// 
+    /// Should consider what the handler actually does (running in context of the event execution) or if some work could be dispatched to an async handler (e.g. service bus). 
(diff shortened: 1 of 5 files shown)


src/backend/api/Fusion.Resources.Api/FusionEvents/LineOrgSyncronizationHandler.cs
```diff
@@ -0,0 +1,81 @@
...
+                {
+                    await HandleOrgUnitUpdatedAsync(payloadData);
+                }
+            }
+        }
+
+        private async Task HandleOrgUnitUpdatedAsync(LineOrgEventBody payloadData)
+        {            
+            // Need to disable cache 
+            using var disableCacheScope = new Fusion.Integration.DisableCacheScope();   // Need to verify this works..
```
aleklundeq, line 64: regarding the comment here, did this work? Or do we need to check this in CI?

aleklundeq (approved):

src/backend/api/Fusion.Resources.Api/FusionEvents/LineOrgSyncronizationHandler.cs
HansDahle: Will fix comment in next pr. Have confirmed that the cache invalidation is used in the integration lib org resolver.

PR #640, opened 2024-06-04
feat: Clean up requests when org unit is deleted

HansDahle:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
As we get more requests and the organisation is re-organising we gather more and more data which is connected to deleted references, e.g deleted org units.

We should try and delete data that will never be relevant (draft requests) and "release" requests that are connected to non-existing units. That way they are at least known to be unassigned.

Updated persistant event handler to publish an `OrgUnitDeleted` notification and two handlers which do solve seperate functionality. Wanted to divide this into smaller functions to increase readability.

Updated "editor" concept to support SystemAccount, when HttpContext is not present. Using Guid.Empty here so we can easily catch it and process it seperately as it will not resolve to an active profile.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

To verify, one can create a new org unit in line org, with a new sap id. Create different requests to verify functionality. The allocation requests can be "reassigned" using the UI, while the change request must be created in a 
different org unit which has employee, then updated using PATCH /.../:requestNumner.
Then the org unit can be deleted in line org and changes verified / confirmed not affected.

- change request which has been sent - Should not be changed at all
- draft change request - should be deleted
- allocation request - should be unassigned

**Checklist:**
- [ ] ~~Considered automated tests~~
- [x] Considered work items 
- [x] Considered security

[AB#53894](https://statoil-proview.visualstudio.com/787035c2-8cf2-4d73-a83e-bb0e6d937eec/_workitems/edit/53894)

### DIFF
--- a/src/backend/api/Fusion.Resources.Api/FusionEvents/LineOrgSyncronizationHandler.cs
+++ b/src/backend/api/Fusion.Resources.Api/FusionEvents/LineOrgSyncronizationHandler.cs
@@ -49,12 +49,19 @@ namespace Fusion.Resources.Api
                 return;
             }
 
-            if (payloadData.GetChangeType() == LineOrgEventBody.ChangeType.Updated)
+            switch (payloadData.GetChangeType())
             {
-                if (payloadData.Changes?.Any(i => i.EqualsIgnCase("fullDepartment")) == true) 
-                {
-                    await HandleOrgUnitUpdatedAsync(payloadData);
-                }
+                case LineOrgEventBody.ChangeType.Updated:
+                    if (payloadData.Changes?.Any(i => i.EqualsIgnCase("fullDepartment")) == true) 
+                    {
+                        await HandleOrgUnitUpdatedAsync(payloadData);
+                    }
+                    break;
+
+                case LineOrgEventBody.ChangeType.Deleted:
+                    await HandleOrgUnitDeletedAsync(payloadData);
+                    break;
+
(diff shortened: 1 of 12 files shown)


HansDahle:
E.g. to create a new org unit in lineorg, under the FOIT sector in TDI. (not, units are softdeleted, so it cannot be reused)

```
POST https://fusion-s-lineorg-ci.azurewebsites.net/org-units
    
    {
        "sapId": "[MY UNIQUE SAP ID (digits only)]",
        "shortName": "HSD",
        "department": "FOS FOIT HSD",
        "fullDepartment": "TDI OG FOS FOIT HSD",
        "name": "Hans Software Dev",
        "type": "CompetenceCenter",
        "level": 5,
        "levelDescription": "Unit level",
        "parentSapId": "52920106",
        "isPortfolio": false,
        "isDeleted": false,
        "parentPath": {
            "level0": "52164905",
            "level1": "53065254",
            "level2": "53067343",
            "level3": "53071980",
            "level4": "52920106",
            "level5": null,
            "level6": null,
            "level7": null,
            "level8": null
        }
    }
```

Requests can be updated by 
```
PATCH https://resources-api.ci.fusion-dev.net/resources/requests/internal/:requestNumber
{
    "assignedDepartment": "[MY UNIQUE SAP ID (digits only)]"
}
```

PR #645, opened 2024-06-26
feat: Summary AZ Function code setup and DepartmentResourceOwnerSync

Jonathanio123:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Work description**:

Created a class library that contains common logic for the two function apps. It's namespace is `Fusion.Resources.Functions.Common`, I feel like there is a better name for it, but cant think of one.

Added DepartmentResourceOwnerSync timer function that retrieves DepartmentResourceOwners for LineOrg and sends them to the new Summary API. It does not put anything on the buss as of now, thinking of adding that in a separate PR.

There's also the question of wheter the summary app should have it's own application registration. As of now it uses resources' registration. But this result in minor code changes for this PR.

**The `department-resource-owner-sync` is disabled because the Summary API still isn't merged in / being worked on**

Link to workitem: US [AB#51132](https://statoil-proview.visualstudio.com/787035c2-8cf2-4d73-a83e-bb0e6d937eec/_workitems/edit/51132), task [AB#54679](https://statoil-proview.visualstudio.com/787035c2-8cf2-4d73-a83e-bb0e6d937eec/_workitems/edit/54679)

**Test description**:

- [ ] Can be tested
- [x] Local tests are passing

**Checklist**:

- [x] Considered work items
- [x] Considered security


src/Fusion.Resources.Functions.Common/ApiClients/SummaryApiClient.cs
```diff
@@ -0,0 +1,35 @@
...
+            MaxDegreeOfParallelism = 10,
+        };
+
+        await Parallel.ForEachAsync(departments, parallelOptions, async (ownerDepartments, token) =>
+        {
+            var body = new JsonContent(JsonSerializer.Serialize(ownerDepartments));
+
+            // Error logging is handled by http middleware => FunctionHttpMessageHandler
+            using var _ = await summaryClient.PutAsync($"departments/{ownerDepartments.DepartmentSapId}", body,
+                cancellationToken);
```
Jonathanio123, line 32: I used Parallel.ForEachAsync to easily limit concurrency, unsure what is preferred practice. Another method could be to chunk the departments and use Task.WhenAll for each chunk.

Alternatively, allow for sending a batch to the endpoint instead

src/Fusion.Resources.Functions.Common/ApiClients/PeopleApiClient.cs
```diff
@@ -1,9 +1,6 @@
-﻿using System;
-using System.Net.Http;
-using System.Threading.Tasks;
-using Fusion.Resources.Functions.Integration;
+﻿using Fusion.Resources.Functions.Common.Integration.Http;
 
-namespace Fusion.Resources.Functions.ApiClients;
+namespace Fusion.Resources.Functions.Common.ApiClients;
 
 public class PeopleApiClient : IPeopleApiClient
```
aleklundeq, line 5: I don't think we need PeopleClient at all for this function. If this is something found from the existing function in the resource repo, the use of PeopleApiClient is because it is used in another function

src/Fusion.Resources.Functions.Common/Configuration/IServiceCollectionExtensions.cs
```diff
@@ -41,15 +40,18 @@ public static IServiceCollection AddHttpClients(this IServiceCollection services
             builder.AddOrgClient();
             services.AddOrgApiClient(HttpClientNames.Application.Org);
 
+            builder.AddSummaryClient();
+            services.AddScoped<ISummaryApiClient, SummaryApiClient>();
+
             builder.AddPeopleClient();
```
aleklundeq, line 46: As mentioned in another comment, we don't need PeopleClient as per now.

src/Fusion.Resources.Functions.Common/ApiClients/PeopleApiClient.cs
Jonathanio123: There seems to be one reference to the IPeopleApiClient here: Fusion.Resources.Functions.InternalRequestsFunctions:106

Jonathanio123: The idea behind the `Fusion.Resources.Functions.Common` project is to be able to reuse what was originally "resources function" code for the new "summary function".

aleklundeq: Yes, that makes sense. In this case we don't need the things related to the other functions. Only the stuff related to Summary notifications should be "merged" over to the new functions.

aleklundeq: Sorry - This was a misread from my part. Looks good!

src/Fusion.Summary.Functions/Functions/DepartmentResourceOwnerSync.cs
```diff
@@ -0,0 +1,56 @@
...
+        var selectedDepartments = departments
+            .Where(d => d.FullDepartment != null).DistinctBy(d => d.SapId).ToList();
+
+        if (!selectedDepartments.Any())
+            throw new Exception("No departments found.");
+
+        var resourceOwners = new List<LineOrgPerson>();
+        foreach (var orgUnitsChunk in selectedDepartments.Chunk(10))
+        {
+            var chunkedResourceOwners =
```
aleklundeq, line 40: This is something that @larfeq is changing in another pr, but for now this should be ok. Maybe add a todo or something where we change the way we get resourceowners

aleklundeq (approved):

PR #664, opened 2024-08-07
Feat: Summary sync puts department on bus

larfeq:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Work description**:

In addition to save department to the database the worker should also send the department (id) to the bus for the weekly worker to pick up. Look at Summary notification architecture (https://miro.com/app/board/uXjVN7jm0Tg=/).

**Expected result**: The servicebus queue is utilized to enable the weekly worker

Link to workitem: [LINK](https://statoil-proview.visualstudio.com/Fusion%20Resource%20Allocation/_workitems/edit/53776)

**Test description**:

- [ ] Can be tested
- [ ] Local tests are passing

**Checklist**:

- [ ] Considered work items
- [ ] Considered security

### DIFF
--- a/pipelines/templates/deploy-function-template.yml
+++ b/pipelines/templates/deploy-function-template.yml
@@ -61,6 +61,7 @@ steps:
           queues = @{
               provisionPosition = "provision-position"
               scheduledNotificationReportQueue = "scheduled-notification"
+              departmentSummaryWeeklyQueue = "department-summary-weekly-queue"
           }
       }
 
--- a/src/backend/function/Fusion.Resources.Functions/Deployment/function.template.json
+++ b/src/backend/function/Fusion.Resources.Functions/Deployment/function.template.json
@@ -27,7 +27,9 @@
         },
         "queues": {
           "provisionPosition": "provision-position",
-          "scheduledNotificationReportQueue": "scheduled-notification"
+          "scheduledNotificationReportQueue": "scheduled-notification",
+          "departmentSummaryWeeklyQueue": "department-summary-weekly-queue"
+          
         }
       }
     },
--- a/src/backend/function/Fusion.Resources.Functions/Deployment/function.template.json
+++ b/src/backend/function/Fusion.Resources.Functions/Deployment/function.template.json
@@ -27,8 +27,7 @@
         },
         "queues": {
           "provisionPosition": "provision-position",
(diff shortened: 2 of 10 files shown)


HansDahle:
Should not the queue that the summary function app uses be the one that creates the queue? It seems like both function deploys create all queues? 

Could the existing function app be reused for summary btw? Do we need to create two? (just a thought :) )

larfeq:
> Should not the queue that the summary function app uses be the one that creates the queue? It seems like both function deploys create all queues?
> 
> Could the existing function app be reused for summary btw? Do we need to create two? (just a thought :) )

I've cleaned up the creation as you commented, good find! 

In regards to your second comment we think we will separate concerns by having the functions grouped in different functions apps.

src/backend/function/Fusion.Resources.Functions/Deployment/function.template.json
```diff
@@ -27,7 +27,9 @@
         },
         "queues": {
           "provisionPosition": "provision-position",
-          "scheduledNotificationReportQueue": "scheduled-notification"
+          "scheduledNotificationReportQueue": "scheduled-notification",
+          "departmentSummaryWeeklyQueue": "department-summary-weekly-queue"
```
Jonathanio123, line 31: Shouldnt this be added to the summary function template?
https://github.com/equinor/fusion-app-resources/blob/5ddaa3cdc0c66b244e8f18e8d2d35d9c398bc338/src/Fusion.Summary.Functions/Deployment/function.template.json

pipelines/templates/deploy-function-template.yml
```diff
@@ -61,6 +61,7 @@ steps:
           queues = @{
               provisionPosition = "provision-position"
               scheduledNotificationReportQueue = "scheduled-notification"
+              departmentSummaryWeeklyQueue = "department-summary-weekly-queue"
```
Jonathanio123, line 64: Shouldn't this be added to a new `deploy-summary-function-template`?

Jonathanio123: Will change when setting up summary function pipelines

Jonathanio123 (approved):

pipelines/templates/deploy-function-template.yml
larfeq: We fix this in a later task

PR #687, opened 2024-09-02
chore: Department filter configuration for az function

Jonathanio123:
- [x] New feature
- [ ] Bug fix
- [ ] High impact

**Description of work:**
[AB#55979](https://statoil-proview.visualstudio.com/787035c2-8cf2-4d73-a83e-bb0e6d937eec/_workitems/edit/55979)

This adds a new setting to the function specifying what departments should be synced and receive reports. The configuration value is a comma separated string. So to only allow departments that contain PRD and TDI => "PRD,TDI".

White space is trimmed and empty values are removed

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security

### DIFF
--- a/src/Fusion.Summary.Functions/Functions/DepartmentResourceOwnerSync.cs
+++ b/src/Fusion.Summary.Functions/Functions/DepartmentResourceOwnerSync.cs
@@ -53,7 +53,7 @@ public class DepartmentResourceOwnerSync
     {
         _serviceBusConnectionString = configuration["AzureWebJobsServiceBus"];
         _weeklySummaryQueueName = configuration["department_summary_weekly_queue"];
-        _departmentFilter = configuration["departmentFilter"]?.Split(',', StringSplitOptions.RemoveEmptyEntries) ?? [];
+        _departmentFilter = configuration["departmentFilter"]?.Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries) ?? ["PRD"];
 
         var client = new ServiceBusClient(_serviceBusConnectionString);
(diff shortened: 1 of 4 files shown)


src/Fusion.Summary.Functions/Functions/DepartmentResourceOwnerSync.cs
```diff
@@ -52,6 +53,7 @@ public async Task RunAsync(
     {
         _serviceBusConnectionString = configuration["AzureWebJobsServiceBus"];
         _weeklySummaryQueueName = configuration["department_summary_weekly_queue"];
+        _departmentFilter = configuration["departmentFilter"]?.Split(',', StringSplitOptions.RemoveEmptyEntries) ?? [];
```
larfeq, line 56: Do we need to clean up white spaces?

Jonathanio123: Maybe something like this would be better? 
```suggestion
        _departmentFilter = configuration["departmentFilter"]?.Split(',').Where(f => !string.IsNullOrWhiteSpace(f)).ToArray() ?? [];
```

If one accidentally adds an extra ', ' it will be removed. Trying to safeguard against accidentally including almost every department due to an empty white space. Maybe adding PRD to the null fallback could also be good idea

Jonathanio123: Pushed a change, now it will also trim values. Think I misunderstood how the department strings work

larfeq (approved):

PR #689, opened 2024-09-03
chore: Summary cleaup

Jonathanio123:
Transitioned to ProblemDetails, updated status codes for endpoints, removed todos, added some extra tests. Changed time trigger for weekly-department-recipients-sync to only run on monday instead of every day.

### DIFF
--- a/src/Fusion.Summary.Api/Controllers/DepartmentsController.cs
+++ b/src/Fusion.Summary.Api/Controllers/DepartmentsController.cs
@@ -6,29 +6,22 @@ using Fusion.Summary.Api.Controllers.ApiModels;
 using Fusion.Summary.Api.Controllers.Requests;
 using Fusion.Summary.Api.Domain.Commands;
 using Fusion.Summary.Api.Domain.Queries;
+using Microsoft.ApplicationInsights.AspNetCore.Extensions;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using HttpRequestExtensions = Microsoft.ApplicationInsights.AspNetCore.Extensions.HttpRequestExtensions;
 
 namespace Fusion.Summary.Api.Controllers;
 
-/// <summary>
-/// TODO: Add summary
-/// </summary>
 [ApiVersion("1.0")]
 [Authorize]
 [ApiController]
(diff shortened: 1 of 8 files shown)


src/Fusion.Summary.Api/Controllers/DepartmentsController.cs
```diff
@@ -43,31 +36,19 @@ public async Task<IActionResult> GetDepartmentsV1()
...
-        if (departments.Length == 0)
-            return NotFound();
-        else
-        {
-            foreach (var d in departments) ret.Add(ApiDepartment.FromQueryDepartment(d));
-        }
+        var apiDepartments = departments.Select(ApiDepartment.FromQueryDepartment);
 
-        return Ok(ret);
+        return Ok(apiDepartments);
```
larfeq, line 43: If list is empty, should we not return no content or is that reserved when nothing (null) is "found"? The endpoint only returns 200 OK

Jonathanio123: You're correct that a 404 status code typically indicates that a resource or endpoint is not found. Since the endpoint in question is valid and functioning as expected, returning a 404 would be misleading. 

When an API endpoint is accessed and successfully processes the request but finds no content to return, it's more appropriate to return a 204 No Content status code if you want to explicitly indicate that the operation was successful but there's no content to send back. However, returning a 200 OK with an empty array is also acceptable and can be seen as an indication that the request was successfully processed and the expected response is an empty list. This can be particularly useful when the client expects a list and can handle an empty list as a normal response.

src/Fusion.Summary.Api/Controllers/DepartmentsController.cs
```diff
@@ -43,31 +36,19 @@ public async Task<IActionResult> GetDepartmentsV1()
 
         #endregion Authorization
 
-        var ret = new List<ApiDepartment>();
+        var departments = await DispatchAsync(new GetAllDepartments());
```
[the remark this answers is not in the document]
Jonathanio123: Unsure if I understand, but doesn't the GetAllDepartments query just return everything? Or are you thinking that we should add some filtering to it so that it does not return everything

larfeq (approved):

PR #737, opened 2025-01-22
fix: set isResourceOwner to true if user has any department responsib…

edmondbaloku:
- [ ] New feature
- [x] Bug fix
- [ ] High impact

**Description of work:**

Because of the new structure introduces with Workday, a user is not in the same department as the department the user has responsibility of. This causes the isResourceOwner property to not get set properly.

In addition to cover all cases when a user isResourceOwner, the property is set to true when the user has any department with responsibility

[AB#59299](https://statoil-proview.visualstudio.com/787035c2-8cf2-4d73-a83e-bb0e6d937eec/_workitems/edit/59299)

**Testing:**
- [ ] Can be tested
- [ ] Local tests are passing

**Checklist:**
- [ ] Considered work items 
- [ ] Considered security

### DIFF
--- a/src/Fusion.Summary.Functions/CardBuilder/AdaptiveCardBuilder.cs
+++ b/src/Fusion.Summary.Functions/CardBuilder/AdaptiveCardBuilder.cs
@@ -65,7 +65,7 @@ public class AdaptiveCardBuilder
     }
 
 
-    public AdaptiveCardBuilder AddGrid(string headerText, string subtitleText, IEnumerable<GridColumn> columnsEnumerable, GoToAction? goToAction = null)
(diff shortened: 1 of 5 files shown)


src/backend/api/Fusion.Resources.Domain/Queries/GetResourceOwnerProfile.cs
```diff
@@ -58,7 +58,7 @@ public Handler(ILogger<Handler> logger, IFusionProfileResolver profileResolver,
                 var departmentsWithResponsibility = await ResolveDepartmentsWithResponsibilityAsync(user, userIsManagerFor);
 
                 // Determine if the user is a manager in the department he/she belongs to.
-                var isDepartmentManager = departmentsWithResponsibility.Any(r => r == user.FullDepartment);
+                var isDepartmentManager = departmentsWithResponsibility.Any();
```
larfeq, line 61: Leaders are placed in "one above", can this be validated?

PR #751, opened 2025-03-04
feat: require location in validation for other tasks

andersndm:
- [ ] New feature
- [x] Bug fix
- [ ] High impact

**Description of work:**

Added location requirement to validation of CreatePersonAbsenceRequest and UpdatePersonAbsenceRequest for absences of type `OtherTasks`.

Added the method `LocationWillBeNull` to the `InternalRequestsController` class to check that a location will be set for the following endpoints:
- [HttpPost("/departments/{departmentString}/resources/requests")]
- [HttpPatch("/resources/requests/internal/{requestId}")]
- [HttpPatch("/departments/{departmentString}/resources/requests/{requestId}")]

The logic is a little complicated, it needs to ensure that after changes will have been applied the request will have a valid location. It needs to take into account whether the request already has a position, if a location is being proposed, and that the proposed location isn't null, i.e. removing an existing location. 

Tests were updated to have a location by default to not cause errors in other existing tests, and tests for the changes were added.

Work Item: [AB#61121](https://statoil-proview.visualstudio.com/787035c2-8cf2-4d73-a83e-bb0e6d937eec/_workitems/edit/61121)

See frontend pr: https://github.com/equinor/fusion-resource-allocation-apps/pull/897

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security

### DIFF
--- a/src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
+++ b/src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
@@ -1,4 +1,5 @@
 ﻿﻿using FluentValidation;
+using Fusion.ApiClients.Org;
 using Fusion.AspNetCore.Api;
 using Fusion.AspNetCore.FluentAuthorization;
 using Fusion.AspNetCore.OData;
@@ -16,6 +17,7 @@ using Fusion.Resources.Logic.Requests;
 using Fusion.Resources.Logic.Workflows;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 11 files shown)


larfeq (approved):

src/backend/api/Fusion.Resources.Api/Controllers/Requests/InternalRequestsController.cs
```diff
@@ -440,7 +453,7 @@ public async Task<ActionResult<ApiResourceAllocationRequest>> PatchInternalReque
 
                 if (request.AdditionalNote.HasValue) updateCommand.AdditionalNote = request.AdditionalNote.Value;
                 if (request.AssignedDepartment.HasValue) updateCommand.AssignedDepartment = request.AssignedDepartment.Value;
-                if (request.ProposedChanges.HasValue) updateCommand.ProposedChanges = request.ProposedChanges.Value;
+                if (request.ProposedChanges?.HasValue ?? false) updateCommand.ProposedChanges = request.ProposedChanges.Value;
```
Jonathanio123, line 456: Is this change needed? If yes should it also be done for the other PatchInternalRequest() method

andersndm: I got a warning there, so fixed it. Seems there is no complaint from the other PatchInternalRequest method, which is odd. It doesn't seem to have been an issue in practice, so I would be happy to revert it.
image

andersndm: removed the null checking

Jonathanio123 (approved):

PR #761, opened 2025-03-25
fix: api should not return null for project state

andersndm:
- [ ] New feature
- [x] Bug fix
- [ ] High impact

**Description of work:**

Link to work item: [AB#60951](https://statoil-proview.visualstudio.com/787035c2-8cf2-4d73-a83e-bb0e6d937eec/_workitems/edit/60951)

Multi-pronged approach setting the project state to `"ACTIVE"` if it is null. This is done in the `QueryProject` and `QueryProjectRef` constructors when querying projects, when projects are added to the database, and when projects are synced with org. Over time all null values should be replaced in the database, and the check in the query layer will act as a failsafe.

Created a string extension to set the state correctly, this will also allow for adding checks in the future.

**Testing:**
- [x] Can be tested
- [x] Local tests are passing

**Checklist:**
- [x] Considered work items 
- [x] Considered security

### DIFF
--- a/src/backend/api/Fusion.Resources.Api/FusionEvents/OrgProjectHandler.cs
+++ b/src/backend/api/Fusion.Resources.Api/FusionEvents/OrgProjectHandler.cs
@@ -3,8 +3,6 @@ using Fusion.Integration.Org;
 using Fusion.Resources.Database;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Logging;
-using System.Text.Json;
-using System.Text.Json.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -61,7 +59,7 @@ namespace Fusion.Resources.Api
             {
                 existingProject.Name = orgProject.Name;
                 existingProject.DomainId = orgProject.DomainId;
-                existingProject.State = orgProject.State;
+                existingProject.State = orgProject.State ?? "ACTIVE";
 
                 await db.SaveChangesAsync(cancellationToken);
             }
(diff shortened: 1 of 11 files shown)


larfeq:
This seems to do the trick, but have you evaluated finding 1) where state is not set (and then is null) and 2) fix the records in the database? The concern is that one time in the future null != active for some reason.

andersndm:
@larfeq Yes, you could fix the records in the database, but that is a somewhat temporary fix. If a project is then synced with org the state would just be set to null again and the same problem arises. Ideally null would be an invalid value for project state in org.

You could fix the records and only retain the check for syncing, but then if null != active in future and the string is changed that would apply whenever a project is synced and still overwrite the state.

I don't see so many benefits to changing the records, when there is currently only one place to change if null != active happens in the future.

Caveat: I am not that familiar with the project syncing so I could be missing something here.

Jonathanio123:
Some cases that make this a bit more complicated

Here null is passed
https://github.com/equinor/fusion-app-resources/blob/fix/null-task-status/src/backend/api/Fusion.Resources.Domain/Utils/PeopleSearchUtils.cs#L158-L158

https://github.com/equinor/fusion-app-resources/blob/fix/null-task-status/src/backend/api/Fusion.Resources.Domain/Utils/PeopleSearchUtils.cs#L243-L243

A related project might not be found
https://github.com/equinor/fusion-app-resources/blob/fix/null-task-status/src/backend/api/Fusion.Resources.Domain/Queries/GetTBNPositions.cs#L125-L125

andersndm:
Yes, added null => ACTIVE to where the projectStates are being created. Then the project state being null would indicate it was not found. Removed resolving the state in the QueryProjectRef constructor since we cannot know if the project state is coming from our database or outside sources.

Jonathanio123 (approved):

